PR #160, opened 2022-04-19
Code clean up 1

ch-kr:
This is the first PR tidying up the code in the RMC repo.

Changes impacting multiple scripts:
- Removed `rtype` from docstrings
- Added code to include chi square significance thresholds used to determine breakpoints to output Table global fields
- Added `VEP_VERSION` constants and updated references to full context resource paths to point to gnomAD methods paths
- Removed code specifically accounting for transcript XG (should be fixed for future runs)
- Updated code finalizing RMC results (both first/additional and simultaneous breaks)

Small changes to specific scripts:
- `rmc/resources/grch37/gnomad.py`: Removed `processed_genomes` resource (not used for v2 RMC results)
- `rmc/resources/grch38/ukbb.py`: Deleted (no longer necessary as all UKB exomes are in v4)
- `rmc/resources/resource_utils.py`: Removed `data_source` (no longer necessary)
- `rmc/utils/generic.py`: Added option to remove outlier transcripts from context HT (so these transcripts will be removed at the beginning of the pipeline)


ch-kr:
closing this old PR; will need to add all changes except this point:
> Updated code finalizing RMC results (both first/additional and simultaneous breaks)
in a future PR

PR #198, opened 2022-06-13
MPC fixes

ch-kr:
This PR contains bug fixes and updates to MPC.

Major changes:
- Updated `get_oe_annotation` to account for duplicate loci in input Table (function now annotates RMC section oe using locus, alleles, and transcript information; rmc/utils/missense_badness.py)
- Moved code creating `context_with_oe` resource (and `context_with_oe_dedup` resource to a new function called `create_context_with_oe`; rmc/utils/mpc.py)
- Updated code creating population/pathogenic variants Table to pull Polyphen-2, transcript, and RMC o/e from `context_with_oe` (rmc/utils/mpc.py)
- Created new function `calculate_fitted_scores` to calculate fitted scores using relationships from MPC regression (rmc/utils/mpc.py)
- Updated gnomAD fitted scores HT creation to use `calculate_fitted_scores` rather than `predict` method (rmc/utils/mpc.py)
- Updated `annotate_mpc` to annotate transcript, a list of MPC values, and MPC float (rmc/utils/mpc.py) 

Minor changes:
- Added `context_with_oe_dedup` resource (rmc/resources/basics.py)
- Added`mpc_dedup` resource (rmc/resources/basics.py)
- Removed `polyphen` resource (rmc/resources/basics.py)
- Added `group_rmc_ht_by_section` utility function to get intervals included in RMC transcript subsections (rmc/utils/constraint.py)
- Fixed typo in `annotate-hts` command call (rmc/pipeline/calculate_mpc.py)
- Removed `n-partitions` argument (no longer used; rmc/pipeline/calculate_mpc.py)
- Added argument and code block to calculate MPC on all gnomAD variants (rmc/pipeline/calculate_mpc.py)
- Added overwrite option to `run_regressions` (rmc/utils/mpc.py)

### DIFF
--- a/rmc/resources/grch37/reference_data.py
+++ b/rmc/resources/grch37/reference_data.py
@@ -22,7 +22,7 @@ Used to calculate the cumulative observed and expected missense values per locus
 generate regional missense constraint results.
 
 NOTE: This resource is created with `process_vep`, which now filters to non-outlier transcripts by default.
-However, for v2, this resource contains *all* transcripts (including outliers).
+However, for v2, this resource contains *all* canonical transcripts (including outliers).
 """
 
 gene_model = TableResource(path=f"{RESOURCE_PREFIX}/GRCh37/browser/b37_transcripts.ht")
--- a/rmc/resources/grch37/reference_data.py
+++ b/rmc/resources/grch37/reference_data.py
@@ -6,7 +6,7 @@ from rmc.resources.resource_utils import RESOURCE_PREFIX
 
 
 ## Reference genome related resources
-processed_context = VersionedTableResource(
+filtered_context = VersionedTableResource(
     default_version="v1",
     versions={
         "v1": TableResource(
--- a/rmc/resources/grch37/reference_data.py
+++ b/rmc/resources/grch37/reference_data.py
@@ -20,6 +20,9 @@ probability of mutation for each variant, CpG status, gnomAD exome coverage, and
 
 Used to calculate the cumulative observed and expected missense values per locus and
 generate regional missense constraint results.
+
+NOTE: This resource is created with `process_vep`, which now filters to non-outlier transcripts by default.
+However, for v2, this resource contains *all* transcripts (including outliers).
 """
 
 gene_model = TableResource(path=f"{RESOURCE_PREFIX}/GRCh37/browser/b37_transcripts.ht")
--- a/rmc/utils/generic.py
+++ b/rmc/utils/generic.py
@@ -260,7 +260,7 @@ def get_exome_bases(build: str) -> int:
     )
 
     logger.info("Removing outlier transcripts...")
-    outlier_transcripts = get_constraint_transcripts(outlier=True)
+    outlier_transcripts = get_constraint_transcripts(outlier=False)
     ht = ht.transmute(transcript_consequences=ht.vep.transcript_consequences)
     ht = ht.explode(ht.transcript_consequences)
     ht = ht.filter(
@@ -375,6 +375,15 @@ def process_vep(ht: hl.Table, filter_csq: bool = False, csq: str = None) -> hl.T
     logger.info("Filtering to canonical transcripts...")
     ht = fast_filter_vep(
         ht, vep_root="vep", syn=False, canonical=True, filter_empty=True
+    )
+
+    logger.info("Filtering to non-outlier transcripts...")
+    # Keep transcripts used in LoF constraint only (remove all other outlier transcripts)
+    constraint_transcripts = get_constraint_transcripts(outlier=True)
+    ht = ht.transmute(transcript_consequences=ht.vep.transcript_consequences)
+    ht = ht.explode(ht.transcript_consequences)
+    ht = ht.filter(
+        constraint_transcripts.contains(ht.transcript_consequences.transcript_id)
     )
 
     logger.info("Annotating HT with most severe consequence...")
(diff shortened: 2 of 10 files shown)


rmc/utils/mpc.py
```diff
@@ -688,6 +749,19 @@ def annotate_mpc(
         hl.tstr
     ), "'locus' must be a LocusExpression, and 'alleles' must be an array of strings!"
 
-    mpc_ht = mpc_release.ht()
-    ht = ht.annotate(mpc=mpc_ht[ht.key].mpc)
+    if "transcript" not in ht.row or add_transcript_annotation:
```
lw453, line 752: Just a note for later that this does not allow annotation with all MPC values for duplicate variants, in case that is something we want to do in the future.

rmc/utils/mpc.py
```diff
@@ -688,6 +749,19 @@ def annotate_mpc(
...
+        assert file_exists(
+            context_with_oe_dedup.path
+        ), "De-duplicated context with OE resource does not exist!"
+        context_ht = context_with_oe_dedup.ht()
+        ht = ht.annotate(transcript=context_ht[ht.key].transcript)
+
+    if not file_exists(mpc_release_dedup.path):
```
lw453, line 759: Wouldn't it make more sense for lines 759-763 to be in `create_mpc_release_ht`?

rmc/utils/mpc.py
```diff
@@ -240,22 +315,31 @@ def prepare_pop_path_ht(
...
+        f"{temp_path}/joint_clinvar_gnomad_transcript.ht", overwrite=True
+    )
 
     logger.info(
-        "Adding PolyPhen-2, codon, and transcript annotations to joint ClinVar/gnomAD HT..."
+        "Getting PolyPhen-2, codon, and regional missense constraint missense o/e annotations..."
     )
+    # Get Polyphen, codon, o/e, amino acid,
+    context_ht = context_with_oe.ht()
     ht = ht.annotate(**context_ht[ht.key])
```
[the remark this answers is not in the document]
ch-kr: ah I forgot to add the code here also keying context with oe with transcript. how would you want to dedup the other annotations? happy to change this again if you think that makes more sense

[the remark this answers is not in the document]
lw453: Adding key on transcript would fix it!

ch-kr:
non-MPC questions/responses:

> Line 275 in generic.py: Should be greater than or equal to not less than.

I changed this to say 'Removing positions' to be clearer

> In process_vep, variants are filtered to canonical transcripts using fast_filter_vep rather than get_constraint_transcripts . Is there a reason why? I think most other instances use get_constraint_transcripts , except get_exome_bases which uses both (could you just use get_constraint_transcripts there?). I'm just wondering if there could be any inconsistencies if we use two different methods.

`fast_filter_vep` filters the `transcript_consequences` subfield within the (gigantic) `vep` annotation. it isn't strictly necessary, but it does make the downstream `explode` smaller.
I've also updated `process_vep` to filter to non-outlier transcripts but have kept the code in  `get_exome_bases` as is (removing outlier transcripts rather than filtering to only non-outlier transcripts). this function is used to determine (roughly) the total number of bases in the exome, which is why I don't want to filter to a small set of transcripts. 

> In line 83 in regional_constraint.py, you use processed_exomes from the gnomAD LOF resources to create filtered_exomes . I think in all other cases when you are manipulating the gnomAD exome data, you use public_release - why is it different here?

haha, this is just because I forgot I used the `processed_exomes` resource (and I lost its location temporarily after the file paths changed). this is something to clean up later; there's no real reason to use `processed_exomes` (we should change to manipulating `public_release` exclusively)

> It seems like add_obs_annotation could use filtered_exomes instead of redoing the loading and filtering of gnomAD?

I updated this function to use `filtered_exomes` if `filter_csq` is set only 

> I think processed_context could have a better name - it filters down to variants that are rare or not seen in gnomAD which isn't clear from the name

I copied this naming from the LoF repo but would love any suggestions!

rmc/utils/generic.py
```diff
@@ -377,6 +377,15 @@ def process_vep(ht: hl.Table, filter_csq: bool = False, csq: str = None) -> hl.T
         ht, vep_root="vep", syn=False, canonical=True, filter_empty=True
     )
 
+    logger.info("Filtering to non-outlier transcripts...")
```
lw453, line 380: This additional filtering step for non-outlier transcripts should be reflected in the description of the resources that are generated with `process_vep` i.e. `filtered_exomes` and `processed_context`.

rmc/utils/generic.py
```diff
@@ -260,7 +260,7 @@ def get_exome_bases(build: str) -> int:
     )
 
     logger.info("Removing outlier transcripts...")
-    outlier_transcripts = get_constraint_transcripts(outlier=True)
+    outlier_transcripts = get_constraint_transcripts(outlier=False)
```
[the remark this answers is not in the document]
ch-kr: oops probably accidentally

rmc/resources/grch37/reference_data.py
```diff
@@ -20,6 +20,9 @@
 
 Used to calculate the cumulative observed and expected missense values per locus and
 generate regional missense constraint results.
+
+NOTE: This resource is created with `process_vep`, which now filters to non-outlier transcripts by default.
+However, for v2, this resource contains *all* transcripts (including outliers).
```
lw453, line 25:
```suggestion
However, for v2, this resource contains *all* canonical transcripts (including outliers).
```

lw453 (approved):

PR #244, opened 2022-09-23
Flow fix 1

ch-kr:
This PR contains runtime fixes for the updated flow and adds a major simultaneous breaks update.

Minor fixes:

- Make default version for `constraint_prep` 1 (we don't need to regenerate this between freezes; we could but is unlikely)
- Added temp code necessary to run initial single search run for freeze 2 (commit is logged and reverted)
- Fixed missing merge arg in `regional_constraint.py`
- Some import fixes in various scripts (removing imports that are no longer used)

Major update:
- Removed loop logic for simultaneous breaks search and replaced with flatmap as discussed in https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/topic/Recording.20iteration-specific.20info.20in.20loop

### DIFF
--- a/rmc/pipeline/two_breaks/run_batches.py
+++ b/rmc/pipeline/two_breaks/run_batches.py
@@ -35,7 +35,6 @@ from rmc.resources.rmc import (
     grouped_single_no_break_ht_path,
     simul_sections_split_by_len_path,
     simul_search_round_bucket_path,
-    single_search_round_ht_path,
 )
 from rmc.slack_creds import slack_token
 from rmc.utils.simultaneous_breaks import check_for_successful_sections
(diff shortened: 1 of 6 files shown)


rmc/pipeline/two_breaks/run_batches.py
```diff
@@ -210,147 +208,82 @@ def search_for_two_breaks(
...
+                    -1,
                 ),
-            ),
+            )
         )
-
+    )
+    # Filter ArrayExpression to only keep chi square values that are at least
+    # some minimum value (to shorten this array and save on storage)
+    break_values_expr.filter(lambda x: x.chisq >= min_chisq_threshold)
```
lw453, line 263:
```suggestion
    break_values_expr = break_values_expr.filter(lambda x: x.chisq >= min_chisq_threshold)
```
And same in `simultaneous_breaks.py`

lw453 (approved):

PR #249, opened 2022-10-07
Bugs found at runtime for simul breaks

ch-kr:
This PR fixes bugs found when running simul breaks search round 1

`rmc/pipeline/regional_constraint.py`: some BLACK reformatting

`rmc/pipeline/two_breaks/merge_hts.py`:
- removed key field from `ANNOTATIONS` and updated line checking length
- added options required to read from requester-pays buckets
- added change to make sure only HTs with results get merged

`rmc/pipeline/two_breaks/run_batches.py`:

- added options required to read from requester-pays buckets
- added missing required imports
- added changes to write all chi square HTs

`rmc/pipeline/two_breaks/run_batches_dataproc.py`:
- added code to avoid specifying sections to run on command line
- added code to make sure TTN was not being included

`rmc/utils/simultaneous_breaks.py`:
- added changes to write all chi square HTs

### DIFF
--- a/rmc/pipeline/two_breaks/run_batches_dataproc.py
+++ b/rmc/pipeline/two_breaks/run_batches_dataproc.py
@@ -42,9 +42,6 @@ def main(args):
             log=f"/round{args.search_num}search_for_two_breaks_run_batches_dataproc.log",
             tmp_dir=TEMP_PATH_WITH_DEL,
         )
-        save_chisq_ht = False
-        if args.search_num == 1 and not args.is_rescue:
-            save_chisq_ht = True
 
         if args.run_ttn:
             section_groups = [[args.ttn_id]]
@@ -101,7 +98,6 @@ def main(args):
                 chisq_threshold=args.chisq_threshold,
                 split_list_len=args.split_list_len,
                 read_if_exists=args.read_if_exists,
-                save_chisq_ht=save_chisq_ht,
             )
 
     finally:
--- a/rmc/pipeline/two_breaks/run_batches_dataproc.py
+++ b/rmc/pipeline/two_breaks/run_batches_dataproc.py
@@ -48,7 +48,8 @@ def main(args):
 
         if args.run_ttn:
             section_groups = [[args.ttn_id]]
-        else:
+
+        if args.run_sections_over_threshold:
             sections_to_run = list(
                 hl.eval(
                     hl.experimental.read_expression(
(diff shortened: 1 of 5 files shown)


rmc/utils/simultaneous_breaks.py
```diff
@@ -410,6 +418,7 @@ def process_section_group(
 
     :param str ht_path: Path to input Table (Table written using `group_no_single_break_found_ht`).
     :param List[str] section_group: List of transcripts or transcript sections to process.
+    :param count: Which transcript group is being run (based on counter generated in `main`).
```
lw453, line 421:
```suggestion
    :param count: Which transcript or transcript section group is being run (based on counter generated in `main`).
```
And the same for all other instances of this

rmc/utils/simultaneous_breaks.py
```diff
@@ -383,7 +385,12 @@ def search_for_two_breaks(
     )
     if save_chisq_ht:
         group_ht = group_ht.checkpoint(
-            f"{SIMUL_BREAK_TEMP_PATH}/temp_chisq_over_threshold.ht", overwrite=True
+            f"{SIMUL_BREAK_TEMP_PATH}/dataproc_temp_chisq_group{count}.ht",
+            overwrite=True,
+        )
+    else:
+        group_ht = group_ht.checkpoint(
+            f"{TEMP_PATH_WITH_DEL}/dataproc_temp_chisq.ht", overwrite=True
```
lw453, line 393: If we are running multiple jobs with save_chisq_ht=False, this could break since they could overwrite each other while running, right? In the batch script, this still has the counter on it

rmc/pipeline/two_breaks/run_batches_dataproc.py
```diff
@@ -145,8 +161,9 @@ def main(args):
     )
     section_ids = parser.add_mutually_exclusive_group()
     section_ids.add_argument(
-        "--sections-to-run",
-        help="Comma separated list of transcripts/transcript sections to run.",
+        "--run-sections-over-threshold",
```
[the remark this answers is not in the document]
ch-kr: not anymore! I'll add it back in to make it more clear that TTN runs on its own, though (which isn't clear from the code as it is now)

rmc/utils/simultaneous_breaks.py
ch-kr: that's true! I only typically spin up one cluster at a time to run the dataproc jobs, though, so this isn't currently an issue

lw453 (approved):

PR #253, opened 2022-11-15
Cumulative obs fix

lw453:
Added changes to `constraint.py` to fix the error in the `cumulative_obs` field computed during the single break search documented [here](https://github.com/broadinstitute/rmc_production/issues/17).


ch-kr (approved):
LGTM, thanks for adding!

PR #264, opened 2023-02-06
Fix log10 > ln

ch-kr:
Also add freezes 5-7 to freezes list

### DIFF
--- a/rmc/pipeline/two_breaks/run_batches.py
+++ b/rmc/pipeline/two_breaks/run_batches.py
@@ -161,7 +161,7 @@ def get_dpois_expr(
     ],
 ) -> hl.expr.StructExpression:
     """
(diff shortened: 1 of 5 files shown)


rmc/pipeline/two_breaks/run_batches.py
```diff
@@ -172,10 +172,8 @@ def get_dpois_expr(
     :return: log10 of the probability under Poisson model.
```
lw453, line 172:
```suggestion
    :return: log of the probability under Poisson model.
```

lw453: Similar change needed on line 164 (and same for corresponding 2 lines in `constraint.py`)

lw453 (approved):

PR #265, opened 2023-02-06
Reference data update

ch-kr:
This PR includes code used to update reference data (see https://github.com/broadinstitute/rmc_production/issues/54).

Updates:

- Added code used to import new reference data to `rmc/utils/generic.py`
- Moved reference data out of `rmc/resources/basics.py` and into `rmc/resources/reference_data.py`
- Updated ClinVar and de novo reference datasets in `rmc/resources/reference_data.py`
- Fixed imports in `rmc/utils/mpc.py`


rmc/utils/generic.py
```diff
@@ -497,77 +509,194 @@ def get_constraint_transcripts(outlier: bool = True) -> hl.expr.SetExpression:
...
+    kap_ht = hl.import_table(ddd_autism_de_novo_tsv, impute=True)
+    kap_ht = kap_ht.transmute(
+        locus=hl.locus(kap_ht.chrom, kap_ht.pos),
+        alleles=[kap_ht.ref, kap_ht.alt],
+    )
+    kap_ht = kap_ht.filter(hl.is_snp(kap_ht.alleles[0], kap_ht.alleles[1]))
+    kap_ht = kap_ht.group_by("locus", "alleles").aggregate(
+        study=hl.agg.collect(kap_ht.study),
+        case_control=hl.agg.collect(kap_ht.case_control),
+    )
```
lw453, line 678: I don't see where a filter was applied here to just individuals with DD/ID?

rmc/utils/generic.py
```diff
@@ -497,77 +509,194 @@ def get_constraint_transcripts(outlier: bool = True) -> hl.expr.SetExpression:
...
+    logger.info(
+        "%i variants failed to liftover, and %i variants flipped strands",
+        liftover_stats.liftover_fail,
+        liftover_stats.flip_strand,
+    )
+    fu_ht = fu_ht.filter(
+        hl.is_defined(fu_ht.new_locus) & ~fu_ht.new_locus.is_negative_strand
+    )
+    fu_ht = fu_ht.key_by(locus=fu_ht.new_locus.result, alleles=fu_ht.alleles)
+    fu_ht = fu_ht.group_by("locus", "alleles").aggregate(
```
lw453, line 650: We shouldn't be grouping these DNM tables by locus + allele as it would discard recurrence information, we just need to retain the phenotype labels and then concatenate the tables

lw453 (changes requested):

rmc/utils/generic.py
ch-kr: I probably forgot to copy it from the notebook, will add

rmc/utils/generic.py
ch-kr: yes -- I did it this way because it was easier to store this smaller, de-duplicated version and simply explode on the sample set afterwards
image

rmc/resources/reference_data.py
```diff
@@ -150,18 +157,19 @@
 and phenotypic context of autism (2022)
 """
 
-de_novo = TableResource(
-    path=f"{RESOURCE_BUILD_PREFIX}/reference_data/ht/ddd_autism_de_novo.ht",
+# TODO: Move HT at `ddd_autism_de_novo.ht` to this path
+ndd_de_novo = TableResource(
```
lw453, line 161: Some references to `de_novo` in `calculate_mpc.py` and `constraint.py` have not yet been switched over to `ndd_de_novo`

rmc/utils/generic.py
```diff
@@ -27,22 +27,21 @@
...
-    clinvar_plp_trip_mis,
-    de_novo,
-    haplo_genes_he,
-    triplo_genes_he,
+    clinvar_plp_mis_haplo,
+    clinvar_plp_mis_triplo,
+    haplo_genes_path,
+    ndd_de_novo,
+    ndd_de_novo_2020_tsv_path,
+    triplo_genes_path,
```
lw453, line 39: Just noticed these aren't in alphabetical order

rmc/utils/generic.py
```diff
@@ -565,36 +568,44 @@ def import_clinvar(overwrite: bool, missense_str: str = MISSENSE) -> None:
...
 
         logger.info("Filtering to variants in haploinsufficient genes...")
-        if not file_exists(haplo_genes):
+        if not file_exists(haplo_genes_path):
             import_dosage(overwrite)
-        hi_genes = hl.experimental.read_expression(haplo_genes_he)
+        hi_genes = hl.experimental.read_expression(haplo_genes_path)
         haplo_ht = ht.filter(hi_genes.contains(ht.gene))
-        haplo_ht = haplo_ht.checkpoint(clinvar_plp_hi_mis.path, overwrite=overwrite)
+        haplo_ht = haplo_ht.checkpoint(clinvar_plp_mis_haplo.path, overwrite=overwrite)
```
lw453, line 595:
```suggestion
        haplo_ht = haplo_ht.checkpoint(
                clinvar_plp_mis_haplo.path,
                _read_if_exists=not overwrite,
                overwrite=overwrite,
        )
```

rmc/utils/generic.py
```diff
@@ -565,36 +568,44 @@ def import_clinvar(overwrite: bool, missense_str: str = MISSENSE) -> None:
...
         )
 
         logger.info("Filtering to variants in triplosensitive genes...")
-        triplo_genes = hl.experimental.read_expression(triplo_genes_he)
+        triplo_genes = hl.experimental.read_expression(triplo_genes_path)
         triplo_ht = ht.filter(triplo_genes.contains(ht.gene))
-        triplo_ht = triplo_ht.checkpoint(clinvar_plp_trip_mis.path, overwrite=overwrite)
+        triplo_ht = triplo_ht.checkpoint(
+            clinvar_plp_mis_triplo.path, overwrite=overwrite
+        )
```
lw453, line 605:
```suggestion
        triplo_ht = triplo_ht.checkpoint(
            clinvar_plp_mis_triplo.path,
            _read_if_exists=not overwrite,
            overwrite=overwrite,
        )
```

rmc/utils/generic.py
ch-kr: that's a good point, I'll switch to a `collect`

rmc/utils/generic.py
ch-kr: good point! these will also be automatically fixed in the future when https://github.com/broadinstitute/regional_missense_constraint/pull/270 is merged

rmc/utils/generic.py
lw453: I'm looking at this again and I think I'm ok with the collect aggregation, I think we can even slim this down more and remove `study` and `sample` and just leave the phenotype info ("DD", "Proband", "Sibling"). I think we should change "Proband" > "ASD" and "Sibling" > "Control" though for downstream ease of use.

rmc/utils/generic.py
```diff
@@ -497,77 +508,220 @@ def get_constraint_transcripts(outlier: bool = True) -> hl.expr.SetExpression:
...
+    kap_ht = hl.import_table(ndd_de_novo_2020_tsv_path, impute=True)
+    kap_ht = kap_ht.transmute(
+        locus=hl.locus(kap_ht.chrom, kap_ht.pos),
+        alleles=[kap_ht.ref, kap_ht.alt],
+    )
+    kap_ht = kap_ht.filter(hl.is_snp(kap_ht.alleles[0], kap_ht.alleles[1]))
+    kap_ht = kap_ht.filter(kap_ht.case_control == "DD")
+    kap_ht = kap_ht.group_by("locus", "alleles").aggregate(
+        study=hl.agg.collect(kap_ht.study),
+        case_control=hl.agg.collect(kap_ht.case_control),
```
lw453, line 696:
```suggestion
case_control=hl.agg.collect(kap_ht.case_control)
```

lw453 (approved):

PR #267, opened 2023-02-15
Add pylint to pre-commit

illusional:
Also add `invalid-unary-operand-type` to pylint ignore due to hail operations not always showing up

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,3 +24,7 @@ repos:
         entry: pylint
         language: system
         types: [python]
+        args: [
+            "-rn", # Only display messages
+            "-sn", # Don't display the score
+          ]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,3 +16,11 @@ repos:
     rev: 6.0.0
     hooks:
       - id: pydocstyle
+  # Using system installation of pylint to support checking python module imports
+  - repo: local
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -16,3 +16,11 @@ repos:
...
     hooks:
       - id: pydocstyle
+  # Using system installation of pylint to support checking python module imports
+  - repo: local
+    hooks:
+      - id: pylint
+        name: pylint
+        entry: pylint
+        language: system
+        types: [python]
```
ch-kr, line 26:
```suggestion
        types: [python]
        args:
            [
              "-rn", # Only display messages
              "-sn", # Don't display the score
              "--rcfile=pylintrc", # Link to pylint config file
            ]
```
I took these lines out of the pylint docs; I'd like pylint to point to our pylintrc

ch-kr (commented):
thanks for catching this! one minor comment

.pre-commit-config.yaml
illusional: It automatically finds the `.pylintrc` file (the default location). The report should be disabled by default, but happy to explicitly disable the report / score, added a commit to address this.

ch-kr: cool, I didn't know that -- thanks for letting me know!

ch-kr (approved):
LGTM!

PR #268, opened 2023-02-15
Use pre-commit in CI

illusional:
Keeps up to date with all pre-commit hooks without having to run the tools separately

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,11 +25,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install wheel
-          pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


ch-kr (commented):
this makes sense to me, though I will note I copied this section from https://github.com/broadinstitute/gnomad_methods/blob/main/.github/workflows/ci.yml. is there ever a reason it would be beneficial to run each step separately?

illusional:
If you wanted a reduced set of checks to run in CI to make it speedy (and cheaper if it's running HEAPs of linting), or wanted to lint extra things before a bigger release you might run them manually, but I'd argue they're already pretty quick.

ch-kr (approved):

PR #269, opened 2023-02-15
Address MYPY suggestions

illusional:
Only minor type annotations required.


ch-kr (approved):
LGTM, thanks for fixing!

PR #270, opened 2023-02-15
Add mypy to pre-commit

illusional:
Feel free to close if not desired, but we find it handy internally to use mypy as it's better at catching type errors.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     hooks:
       - id: pydocstyle
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.961
+    rev: v1.0.1
     hooks:
       - id: mypy
         args:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,3 +30,9 @@ repos:
             --install-types,
             --non-interactive,
           ]
+    - repo: https://github.com/pycqa/isort
+      rev: 5.10.1 # This should be kept in sync with the version in requirements-dev.in
+      hooks:
+        - id: isort
+          args: ["--profile", "black", "--filter-files"]
+          files: rmc
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,3 +16,17 @@ repos:
     rev: 6.0.0
     hooks:
       - id: pydocstyle
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.961
+    hooks:
+      - id: mypy
+        args:
+          [
+            --pretty,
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -16,3 +16,17 @@ repos:
     rev: 6.0.0
     hooks:
       - id: pydocstyle
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.961
```
ch-kr, line 21: why not use the most recent version (1.0.0)?

.pre-commit-config.yaml
```diff
@@ -16,3 +16,17 @@ repos:
...
+      - id: mypy
+        args:
+          [
+            --pretty,
+            --show-error-codes,
+            --no-strict-optional,
+            --ignore-missing-imports,
+            --install-types,
+            --non-interactive,
+          ]
```
ch-kr, line 32: I'm going to sneak this into this PR as well (another item I've been meaning to add):
```suggestion
          ]
    - repo: https://github.com/pycqa/isort
      rev: 5.10.1 # This should be kept in sync with the version in requirements-dev.in
      hooks:
        - id: isort
          args: ["--profile", "black", "--filter-files"]
          files: rmc
```

ch-kr (commented):
thanks for adding this -- this has actually been on my todo list for some time!

ch-kr:
basic question, wouldn't we also need to add `mypy` to `requirements-dev.in`? (and `isort`)

illusional:
The `pre-commit` library will keep everything installed and correct version itself (except pylint which I listed as `local`, because it works better with packages that way), so no need to add them explicitly to requirements-dev.

illusional:
(Just noting that mypy / isort aren't being run in CI without merging #268 first.

ch-kr:
Yes, I just merged that other PR. Thank you for adding it, and for adding isort! My only remainining question here is around the version; it seems like the most recent `mypy` version is 1.0.0. Is there a reason you prefer v0.961?

ch-kr:
pre-commit failures fixed in https://github.com/broadinstitute/regional_missense_constraint/pull/275

.pre-commit-config.yaml
```diff
@@ -16,6 +16,25 @@ repos:
     rev: 6.0.0
     hooks:
       - id: pydocstyle
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.961
```
ch-kr, line 20:
```suggestion
    rev: v1.0.1
```

ch-kr:
closing; duplicated in https://github.com/broadinstitute/regional_missense_constraint/pull/278

PR #278, opened 2023-03-01
mypy and isort

ch-kr:
I botched the formatting in this [PR](https://github.com/broadinstitute/regional_missense_constraint/pull/270/files) when merging with main, so adding mypy and isort here


ch-kr:
Also updated pydocstyle to match version used in gnomAD QC repo and updated python version to be compatible with isort

ch-kr:
also ran isort because it would fix a bunch of scripts; see https://github.com/broadinstitute/regional_missense_constraint/actions/runs/4315411425/jobs/7529751561

PR #281, opened 2023-03-08
Reference data bug fixes

ch-kr:
This PR removes TODOs left in reference data import code (after paths were fixed in https://github.com/broadinstitute/rmc_production/issues/89) and adds minor fixes to the NDD de novo dataset import code

- Removed TODOs around NDD de novo and ClinVar P/LP haplo HT paths
- Added accidentally missed annotations in function to import Fu Data
- Added accidentally missed not (`~`) in import Fu data function
- Fixed sample set join in import de novo function
^I know this last one is ugly, but using [`extend`](https://hail.is/docs/0.2/hail.expr.ArrayExpression.html#hail.expr.ArrayExpression.extend) to concatenate the arrays wasn't working, so I picked an uglier method. Very open to suggestions here

### DIFF
--- a/rmc/utils/generic.py
+++ b/rmc/utils/generic.py
@@ -727,20 +727,13 @@ def import_de_novo_variants(overwrite: bool, n_partitions: int = 5000) -> None:
 
     # Union sample types (DD, ASD, control)
     ht = ht.transmute(
-        role_str=hl.str(",").join(ht.role),
-        case_control_str=hl.str(",").join(ht.case_control),
-    )
-    ht = ht.transmute(
         sample_set=hl.case()
         .when(
-            hl.is_defined(ht.case_control_str) & hl.is_defined(ht.role_str),
-            ht.case_control_str + ht.role_str,
+            hl.is_defined(ht.case_control) & hl.is_defined(ht.role),
+            ht.case_control.extend(ht.role),
         )
-        .when(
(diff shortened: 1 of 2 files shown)


rmc/utils/generic.py
```diff
@@ -702,17 +714,33 @@ def import_de_novo_variants(overwrite: bool, n_partitions: int = 5000) -> None:
...
+            hl.is_defined(ht.case_control_str) & hl.is_defined(ht.role_str),
+            ht.case_control_str + ht.role_str,
+        )
+        .when(
+            hl.is_defined(ht.case_control_str) & hl.is_missing(ht.role_str),
+            ht.case_control_str,
+        )
+        .default(ht.role_str)
+    )
+    ht = ht.transmute(sample_set=ht.sample_set.split(","))
```
lw453, line 745: We should just be able to do this, right? Unless I'm missing something
```
ht1 = ht.transmute(
    sample_set = hl.case()
    .when(hl.is_defined(ht.case_control) & hl.is_defined(ht.role), ht.case_control.extend(ht.role))
    .when(hl.is_defined(ht.case_control), ht.case_control)
    .when(hl.is_defined(ht.role), ht.role)
    .or_missing()
)
```

ch-kr: chatted on slack; adding this because the extend worked for Lily

lw453 (approved):

PR #282, opened 2023-03-10
Rearrange RMC and generic code

lw453:
This PR moves the code to generate the OE-annotated VEP context table into the RMC utils file and moves the code calling it into the `finalize` section of the RMC pipeline file (originally located in the MPC pipeline file). It also moves a few other functions out of the missense badness and MPC utils files and into the RMC or generic utils files where they more appropriately belong.

### DIFF
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -552,13 +552,13 @@ def main(args):
             rmc_ht = rmc_ht.filter(constraint_transcripts.contains(rmc_ht.transcript))
 
             logger.info("Writing out RMC results...")
-            rmc_ht.write(rmc_results.versions[int(args.freeze)].path)
+            rmc_ht.write(rmc_results.versions[args.freeze].path)
 
             logger.info("Getting transcripts without evidence of RMC...")
             create_no_breaks_he(freeze=args.freeze, overwrite=args.overwrite)
 
             logger.info("Creating OE-annotated context table...")
-            create_context_with_oe(freeze=int(args.freeze), overwrite_output=args.overwrite)
+            create_context_with_oe(freeze=args.freeze, overwrite_output=args.overwrite)
 
     finally:
         logger.info("Copying hail log to logging bucket...")
@@ -644,6 +644,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "--freeze",
         help="RMC data freeze number",
+        type=int,
         default=CURRENT_FREEZE,
     )
     parser.add_argument(
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -35,6 +35,7 @@ from rmc.utils.constraint import (
     calculate_exp_per_transcript,
     calculate_observed,
     check_break_search_round_nums,
+    create_context_with_oe,
     create_no_breaks_he,
     get_max_chisq_per_group,
     GROUPINGS,
@@ -555,6 +556,11 @@ def main(args):
 
             logger.info("Getting transcripts without evidence of RMC...")
             create_no_breaks_he(freeze=args.freeze, overwrite=args.overwrite)
+
+            logger.info("Creating OE-annotated context table")
+            create_context_with_oe(
+                overwrite_temp=args.overwrite, overwrite_output=args.overwrite
+            )
 
     finally:
         logger.info("Copying hail log to logging bucket...")
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -557,7 +557,7 @@ def main(args):
             logger.info("Getting transcripts without evidence of RMC...")
             create_no_breaks_he(freeze=args.freeze, overwrite=args.overwrite)
 
-            logger.info("Creating OE-annotated context table")
+            logger.info("Creating OE-annotated context table...")
             create_context_with_oe(freeze=args.freeze, overwrite_output=args.overwrite)
 
     finally:
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -558,7 +558,7 @@ def main(args):
             create_no_breaks_he(freeze=args.freeze, overwrite=args.overwrite)
 
             logger.info("Creating OE-annotated context table...")
-            create_context_with_oe(freeze=args.freeze, overwrite_output=args.overwrite)
+            create_context_with_oe(freeze=int(args.freeze), overwrite_output=args.overwrite)
 
     finally:
         logger.info("Copying hail log to logging bucket...")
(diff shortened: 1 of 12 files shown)


rmc/pipeline/regional_constraint.py
```diff
@@ -556,6 +557,9 @@ def main(args):
             logger.info("Getting transcripts without evidence of RMC...")
             create_no_breaks_he(freeze=args.freeze, overwrite=args.overwrite)
 
+            logger.info("Creating OE-annotated context table...")
+            create_context_with_oe(freeze=args.freeze, overwrite_output=args.overwrite)
```
ch-kr, line 561:
```suggestion
            create_context_with_oe(freeze=int(args.freeze), overwrite_output=args.overwrite)
```

rmc/utils/generic.py
```diff
@@ -237,6 +276,41 @@ def filter_context_using_gnomad(
     return context_ht
 
 
+def get_annotations_from_context_ht_vep(
+    ht: hl.Table, annotations: List[str] = ["polyphen", "sift"]
```
ch-kr, line 280: while we're updating this function, should we remove `sift` here since we no longer compare MPC to sift?

rmc/utils/missense_badness.py
```diff
@@ -140,6 +42,7 @@ def prepare_amino_acid_ht(
         - Add observed and OE annotation
         - Write to `amino_acids_oe` resource path
 
+    :param freeze: RMC freeze number. Default is CURRENT_FREEZE.
```
ch-kr, line 45: this is ordered first in the docstring but last in the args -- either order works for me, but this should probably be consistent

rmc/utils/missense_badness.py
```diff
@@ -28,107 +26,11 @@
...
-    ht = ht.annotate(
-        section_oe=hl.or_missing(
-            hl.len(ht.section_oe) > 0,
-            ht.section_oe[0],
-        ),
-    )
-    return ht.transmute(oe=hl.coalesce(ht.section_oe, ht.transcript_oe))
-
-
 def prepare_amino_acid_ht(
```
ch-kr, line 29: we should update the call to this function to add freeze also (https://github.com/broadinstitute/regional_missense_constraint/blob/main/rmc/pipeline/calculate_missense_badness.py#L32)

rmc/utils/mpc.py
```diff
@@ -363,11 +238,10 @@ def prepare_pop_path_ht(
     logger.info("Getting transcript annotations...")
     # If OE-annotated VEP context HT filtered to missense variants
     # in canonical transcripts does not exist, create it
-    if not file_exists(context_with_oe_dedup.path):
-        create_context_with_oe(
-            overwrite_temp=overwrite_temp,
-            overwrite_output=overwrite_output,
-        )
+    if not file_exists(context_with_oe_dedup.path) or not file_exists(
```
ch-kr, line 241: we should add to the docstring that this function now assumes these resources exist. also, should we add freeze here (and to the function call [here](https://github.com/broadinstitute/regional_missense_constraint/blob/main/rmc/pipeline/calculate_mpc.py#L39))?

ch-kr (changes requested):
thanks for updating these! I have only a few small suggestions

rmc/utils/generic.py
lw453: After discussion, we agreed to keep `sift`

lw453: Also this made me realize that the default for `annotations` isn't noted in this function's docstring (pushing a fix)

rmc/utils/missense_badness.py
lw453: Yep, good catch!

rmc/utils/mpc.py
lw453: Added changes to address both

lw453:
Additional major change made: Freeze arguments added for missense badness and MPC functions. 
Additional minor change made: Moved some temp misbad and MPC resources from `TEMP_PATH` to `TEMP_PATH_WITH_FAST_DEL`.

rmc/resources/rmc.py
```diff
@@ -497,48 +497,55 @@ def no_breaks_he_path(
...
+        for freeze in FREEZES
+    },
+)
+"""
+Table containing "population" and "pathogenic" variants.
 
-    Table is input to MPC (missense badness, polyphen-2, and constraint) calculations.
+Table contains common (AF > 0.001) gnomAD variants ("population") and
+ClinVar pathogenic/likely pathogenic missense variants in haploinsufficient genes
+that cause severe disease ("pathogenic") with defined CADD, BLOSUM, Grantham, missense observed/expected ratios,
```
ch-kr, line 514:
```suggestion
that cause severe disease ('pathogenic') with defined CADD, BLOSUM, Grantham, missense observed/expected ratios,
```

rmc/resources/rmc.py
```diff
@@ -497,48 +497,55 @@ def no_breaks_he_path(
...
-    :return: Path to Table.
-    """
-    # TODO: convert back into TableResource
-    return f"{CURRENT_MPC_PREFIX}/joint_clinvar_gnomad.ht"
+Table is input to MPC (missense badness, polyphen-2, and constraint) calculations.
+"""
 
 
-def mpc_model_pkl_path() -> str:
+def mpc_model_pkl_path(freeze: int = CURRENT_FREEZE) -> str:
```
ch-kr, line 521: do you think we'll generate multiple models? if we only generate MPC on a single freeze, then we'll know from the MPC results path which RMC freeze it's associated with, so we don't need to add this here (but fine keeping this also)

rmc/utils/mpc.py
```diff
@@ -173,139 +167,21 @@ def import_grantham():
...
     overwrite_temp: bool = False,
     overwrite_output: bool = True,
+    freeze: int = CURRENT_FREEZE,
 ) -> None:
     """
     Prepare Table with 'population' (common gnomAD missense) and 'pathogenic' (ClinVar pathogenic/likely pathogenic missense) variants.
 
     .. note::
-        This function reads in data from a requester-pays bucket and will fail if requester-pays
+        - This function reads in data from a requester-pays bucket and will fail if requester-pays
```
ch-kr, line 181: ooh I missed this last time, but the entire RMC bucket is requester-pays now, so we should probably remove this
```suggestion
```

ch-kr (commented):
a few minor suggestions + one question

rmc/resources/rmc.py
```diff
@@ -497,48 +497,55 @@ def no_breaks_he_path(
...
+    default_version=CURRENT_FREEZE,
+    versions={
+        freeze: TableResource(
+            path=f"{MPC_PREFIX}/{CURRENT_GNOMAD_VERSION}/{freeze}/joint_clinvar_gnomad.ht"
+        )
+        for freeze in FREEZES
+    },
+)
+"""
+Table containing "population" and "pathogenic" variants.
```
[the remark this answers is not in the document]
lw453: I just copied this from the previous docstring (when it was a function), happy to change it

rmc/resources/rmc.py
lw453: We discussed this together and I had added these args to MPC and misbad because they were requested for a couple of functions so I had just applied to everything. So we'll leave it for now

PR #283, opened 2023-03-29
Misbad and MPC runtime fixes

ch-kr:
Small fixes for bugs at runtime, also added functionality to support MPC model that uses all interactions


lw453:
We also need to update the `clinvar` resource to point to `gs://regional_missense_constraint/resources/GRCh37/reference_data/ht/clinvar.GRCh37.ht/` instead of the seqr table and document that it corresponds to the 20230305 ClinVar release.

Ruchit10 (approved):

PR #286, opened 2023-05-02
MPC variants update

ch-kr:
Major changes:
- Update `prepare_pop_path_ht` function to filter gnomAD public release to high quality, common variants (previously only filtered gnomAD public release HT on AF)
- Update `prepare_pop_path_ht` function to filter joint ClinVar/gnomAD HT to set of transcripts to keep (previously kept variants in all transcripts)

Minor changes:
- Created `get_gnomad_public_release` function (moved code getting gnomAD public release HT and annotating with coverage to a function)
- Added `filter_to_rare` option to `keep_criteria` function

### DIFF
--- a/rmc/utils/generic.py
+++ b/rmc/utils/generic.py
@@ -371,12 +371,12 @@ def keep_criteria(
     :param cov_threshold: Remove rows at or below this median coverage threshold. Default is 0.
     :param filter_to_rare: Whether to filter to keep rare variants only.
         If True, only variants with AF < `af_threshold` will be kept.
-        If False, only variants with AF > `af_threshold` will be kept.
+        If False, only variants with AF >= `af_threshold` will be kept.
         Default is True.
     :return: Boolean expression used to filter variants.
     """
     af_filter_expr = (
-        (af_expr < af_threshold) if filter_to_rare else (af_expr > af_threshold)
+        (af_expr < af_threshold) if filter_to_rare else (af_expr >= af_threshold)
     )
     return (
         (ac_expr > 0)
--- a/rmc/utils/generic.py
+++ b/rmc/utils/generic.py
@@ -246,7 +246,6 @@ def filter_context_using_gnomad(
     context_ht: hl.Table,
     gnomad_data_type: str = "exomes",
     adj_freq_index: int = 0,
-    filter_context_using_cov: bool = True,
     cov_threshold: int = 0,
 ) -> hl.Table:
     """
@@ -258,10 +257,7 @@ def filter_context_using_gnomad(
         Default is "exomes".
     :param adj_freq_index: Index of frequency array that contains global population filtered calculated on
         high quality (adj) genotypes. Default is 0.
-    :param filter_context_using_cov: Whether to also filter sites in context Table using gnomAD coverage.
-        Default is True.
-    :param cov_threshold: Coverage threshold used to filter context Table if `filter_context_using_cov` is True.
-        Default is 0.
+    :param cov_threshold: Coverage threshold used to filter context Table. Default is 0.
     :return: Filtered VEP context Table.
     """
     gnomad = get_gnomad_public_release(gnomad_data_type, adj_freq_index)
@@ -278,18 +274,6 @@ def filter_context_using_gnomad(
             cov_threshold=cov_threshold,
         )
     )
-
-    # Optionally also filter context HT using gnomAD coverage
-    if filter_context_using_cov:
-        gnomad_cov = coverage(gnomad_data_type).ht()
-        context_ht = context_ht.annotate(
-            gnomad_coverage=gnomad_cov[context_ht.locus].median
-        )
-        context_ht = context_ht.filter(context_ht.gnomad_coverage > cov_threshold)
-        # Drop coverage annotation here for consistency
-        # (This annotation is only added if `filter_context_using_cov` is True)
-        context_ht = context_ht.drop("gnomad_coverage")
-
     return context_ht
 
 
--- a/rmc/utils/generic.py
+++ b/rmc/utils/generic.py
@@ -228,6 +228,8 @@ def get_gnomad_public_release(
     :param gnomad_data_type: gnomAD data type. Used to retrieve public release and coverage resources.
         Must be one of "exomes" or "genomes" (check is done within `public_release`).
         Default is "exomes".
+    :param adj_freq_index: Index of frequency array that contains global population filtered calculated on
+        high quality (adj) genotypes. Default is 0.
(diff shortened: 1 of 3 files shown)


rmc/utils/mpc.py
```diff
@@ -282,14 +304,18 @@ def prepare_pop_path_ht(
         overwrite=overwrite_temp,
     )
 
-    logger.info("Filtering to rows with defined annotations and writing out...")
+    logger.info(
+        "Filtering to rows with defined annotations in transcripts to keep and writing out..."
```
lw453, line 308:
```suggestion
        "Filtering to rows with defined annotations in non-outlier transcripts and writing out..."
```

rmc/utils/generic.py
```diff
@@ -349,23 +370,28 @@ def keep_criteria(
...
-    :param hl.expr.Float64Expression af_expr: Allele frequency (AF) Float64Expression.
-    :param hl.expr.SetExpression filters_expr: Filters SetExpression.
-    :param hl.expr.Int32Expression cov_expr: gnomAD median coverage Int32Expression.
-    :param float af_threshold: Remove rows above this AF threshold. Default is 0.001.
-    :param int cov_threshold: Remove rows below this median coverage threshold. Default is 0.
+    :param ac_expr: Allele count Int32Expression.
+    :param af_expr: Allele frequency (AF) Float64Expression.
+    :param filters_expr: Filters SetExpression.
+    :param cov_expr: gnomAD median coverage Int32Expression.
+    :param af_threshold: Remove rows above this AF threshold. Default is 0.001.
```
lw453, line 384:
```suggestion
    :param af_threshold: AF threshold below which variants are removed. Default is 0.001.
```

rmc/utils/generic.py
```diff
@@ -349,23 +370,28 @@ def keep_criteria(
...
+    :param cov_threshold: Remove rows below this median coverage threshold. Default is 0.
```
lw453, line 385:
```suggestion
    :param cov_threshold: Median coverage threshold below which variants are removed. Default is 0.
```

lw453 (changes requested):

rmc/utils/generic.py
```diff
@@ -259,6 +279,7 @@ def filter_context_using_gnomad(
 
     # Optionally also filter context HT using gnomAD coverage
     if filter_context_using_cov:
```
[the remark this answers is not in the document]
ch-kr: I also can't remember why this is separated out from the rest of the function, since it looks like it gets called once and is set to True? so I'll remove it entirely
image

rmc/utils/generic.py
```diff
@@ -349,23 +370,28 @@ def keep_criteria(
...
+    :param af_expr: Allele frequency (AF) Float64Expression.
+    :param filters_expr: Filters SetExpression.
+    :param cov_expr: gnomAD median coverage Int32Expression.
+    :param af_threshold: Remove rows above this AF threshold. Default is 0.001.
+    :param cov_threshold: Remove rows below this median coverage threshold. Default is 0.
+    :param filter_to_rare: Whether to filter to keep rare variants only. Default is True.
     :return: Boolean expression used to filter variants.
     """
+    af_filter_expr = (
+        (af_expr < af_threshold) if filter_to_rare else (af_expr > af_threshold)
```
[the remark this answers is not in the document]
ch-kr: as in, <= or >=? we only filter with the < and > if I remember correctly (but could use a fact check here)

rmc/utils/generic.py
ch-kr: GitHub isn't letting me apply this suggestion (and the one above), but I think I prefer the active vs passive voice here

rmc/utils/generic.py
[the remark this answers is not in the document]
ch-kr: sounds good to me, I'll update MPC (agree that we don't want to redo RMC if possible)

rmc/utils/generic.py
```diff
@@ -254,12 +255,9 @@ def filter_context_using_gnomad(
     :param gnomad_data_type: gnomAD data type. Used to retrieve public release and coverage resources.
         Must be one of "exomes" or "genomes" (check is done within `public_release`).
         Default is "exomes".
-    :param adj_freq_index: Index of frequency array that contains global population filtered calculated on
+    :param adj_freq_index: Index of frequency array that contains global population calculated on
         high quality (adj) genotypes. Default is 0.
```
lw453: Phrasing that may be clearer: `Index of array containing allele frequencies calculated on high-quality (adj) genotypes over various population groupings. Default is 0.`

rmc/utils/generic.py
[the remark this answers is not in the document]
lw453: Same phrasing should be used in line 260

lw453: Ok I made this suggestion because the original wording makes me initially think of a boolean (remove or not), but it's understandable nonetheless

rmc/utils/mpc.py
```diff
@@ -189,7 +189,7 @@ def prepare_pop_path_ht(
         Must be one of "exomes" or "genomes" (check is done within `public_release`).
         Default is "exomes".
     :param float af_threshold: Allele frequency cutoff to filter gnomAD public dataset.
-        Variants *above* this threshold will be kept.
+        Variants greater than or equal to this threshold will be kept.
```
lw453, line 192:
```suggestion
        Variants at frequencies greater than or equal to this threshold will be kept.
```

lw453 (changes requested):

rmc/utils/generic.py
ch-kr: ah yes, I'm used to thinking about this as global from QC. will update!

rmc/utils/generic.py
```diff
@@ -228,8 +228,8 @@ def get_gnomad_public_release(
     :param gnomad_data_type: gnomAD data type. Used to retrieve public release and coverage resources.
         Must be one of "exomes" or "genomes" (check is done within `public_release`).
         Default is "exomes".
-    :param adj_freq_index: Index of frequency array that contains global population calculated on
-        high quality (adj) genotypes. Default is 0.
+    :param adj_freq_index: Index of frequency array that contains frequency information calculated on
+        high quality (adj) genotypes across all genetic ancestry groups. Default is 0.
```
lw453, line 232: Just one more suggestion for this wording (and same for the description in `mpc.py`:
```suggestion
    :param adj_freq_index: Index of array that contains allele frequency information calculated on
        high quality (adj) genotypes across genetic ancestry groups. Default is 0.
```

lw453 (approved):

PR #290, opened 2023-06-01
Add blosum and grantham compatibility to calculate_fitted_scores

Ruchit10:
Minor changes:
1. Add compatibility for blosum and grantham scores from the updated regression model to `calculate_fitted_scores`

### DIFF
--- a/rmc/utils/mpc.py
+++ b/rmc/utils/mpc.py
@@ -620,10 +620,10 @@ def calculate_fitted_scores(
         (ht[variable] * mpc_rel_vars[variable]) for variable in variable_dict.keys()
     ]
     interaction_annot_expr = []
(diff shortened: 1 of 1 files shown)


rmc/utils/mpc.py
```diff
@@ -605,22 +621,10 @@ def calculate_fitted_scores(
     ]
     interaction_annot_expr = []
     for variable in interactions_dict.keys():
```
lw453, line 623: I think renaming `variable` to `interaction` would be more clear here — and then rename `var_split` to `variable` on 625/626

lw453 (changes requested):

rmc/utils/mpc.py
Ruchit10: Done

PR #291, opened 2023-06-05
Flexibility updates for training transcript sets for missense badness and MPC

lw453:
This PR adds updates for flexibility in training up different models for missense badness and MPC depending on input training transcript sets and model components. Plus some other miscellaneous updates.

Major changes:

- MPC-related:
    - Addition of capacity to train MPC models on different input training transcript sets (e.g. all training transcripts vs. training transcripts from a specific fold) and model components
        - E.g. bringing in missense badness scores trained on a specific training set in construction of an MPC model for that specific training set
        - Turning MPC-related resource definitions into functions that can return different paths for different input training transcript sets and model components
    - Addition of support for utils functions in the pipeline file `calculate_mpc.py`
    - Addition of capacity to train MPC models using a specific input formula or by trying out models from mix-and-match of a set of input variables/components and keeping the best one.
        - The multiple functionalities previously contained within `run_regressions` have now been separated into `run_glm` (for just fitting a single regression), `get_min_aic_model` (to get the model with the minimum AIC from mix-and-matching an input set of variables) and `run_regressions` (the top-level function).
    - Addition of capacity to annotate MPC scores onto a given Table of input variants using the MPC release HT (containing all VEP context missense variants and their MPC scores) or computing scores from a saved model without using a prewritten release HT. This allows us to try out different models without the burden of writing out the entire length of the context HT for each model.
    - Restructuring of `calculate_fitted_scores` for code efficiency and flexibility for different models
    - Removal of the version of the MPC release Table that has multiple rows for variants in multiple transcripts. Only the "deduplicated" version of this Table is now retained.
        - This necessitated some tinkering with how fitted scores are annotated in `calculate_fitted_scores` i.e. handling on what happens when a variant is located in multiple transcripts.
    - Addition of a suffix label on temporary files (`temp_label`) generated during the process of model generation to avoid conflicting file writes. (This will be further updated in later commits to be converted to a subfolder structure.)
- Missense badness-related:
    - Removal of code related to creation of missense badness models with anything other than training transcripts, i.e. testing transcripts or validation transcripts from a specific fold. This was mistakenly added in previous PRs.
    - Code efficiency improvements in `prepare_amino_acid_ht` and `calculate_misbad`

Minor changes:

- Addressing code from previous PRs:
    - Addition of a temporary save point for an aggregation that had been made for logging purposes on the HC LoF filter. This update means that the aggregation does not have to be performed every time `prepare_amino_acid_ht` is run (e.g., useful when this function is rerun with different training transcript sets).
    - Merging of functions returning paths to training, validation, and test transcripts in `reference_data.py`
    - Removal of the `overwrite_output` parameter which was a vestige from previous experimentation with RMC regions. All output resources are now always written out (`overwrite=True`). This and the points regarding line spacing and documentation should be the only changes made to `constraint.py` and `regional_constraint.py` (code for RMC regions and not missense badness or MPC).
- Addition of subfolder structure to paths for missense badness and MPC resources in `rmc.py`, so that resources corresponding to models trained on a particular transcript set are located in the same folder.
- Formatting/spacing for various lines
- Updates to documentation for various function/argparse arguments
- Small changes to some temp file names

### DIFF
--- a/rmc/pipeline/calculate_mpc.py
+++ b/rmc/pipeline/calculate_mpc.py
@@ -44,6 +44,8 @@ def main(args):
         if args.command == "run-glm":
             hl.init(log="/run_regressions_using_glm.log", tmp_dir=temp_dir)
             run_regressions(
+                use_model_formula=args.use_model_formula,
+                model_formula=args.model_formula,
                 variables=args.variables.split(","),
                 additional_variables=args.extra_variables.split(","),
                 overwrite_temp=args.overwrite_temp,
@@ -67,9 +69,13 @@ def main(args):
                 annotate_mpc(
                     ht=clinvar_plp_mis_haplo.ht(),
                     output_ht_path=f"{mpc_bucket_path}/clinvar_mpc_annot.ht",
+                    temp_label="_clinvar",
+                    model_train_fold=args.model_train_fold,
+                    use_release=args.use_release,
                     overwrite_temp=args.overwrite_temp,
                     freeze=args.freeze,
                 )
+                # TODO: Add support for annotating with models from specific folds (all folds?)
 
             if args.dd:
                 from rmc.resources.reference_data import ndd_de_novo
@@ -80,6 +86,9 @@ def main(args):
                 annotate_mpc(
                     ht=case_ht,
                     output_ht_path=f"{mpc_bucket_path}/dd_case_mpc_annot.ht",
+                    temp_label="_dd_case",
+                    model_train_fold=args.model_train_fold,
+                    use_release=args.use_release,
                     overwrite_temp=args.overwrite_temp,
                     freeze=args.freeze,
                 )
@@ -92,6 +101,9 @@ def main(args):
                 annotate_mpc(
                     ht=control_ht,
                     output_ht_path=f"{mpc_bucket_path}/dd_control_mpc_annot.ht",
+                    temp_label="_dd_control",
+                    model_train_fold=args.model_train_fold,
+                    use_release=args.use_release,
                     overwrite_temp=args.overwrite_temp,
                     freeze=args.freeze,
                 )
@@ -103,6 +115,9 @@ def main(args):
                 annotate_mpc(
                     ht=ht,
                     output_ht_path=f"{mpc_bucket_path}/gnomAD_mpc_annot.ht",
+                    temp_label="_gnomad_exome",
+                    model_train_fold=args.model_train_fold,
+                    use_release=args.use_release,
                     overwrite_temp=args.overwrite_temp,
                     freeze=args.freeze,
                 )
@@ -112,6 +127,9 @@ def main(args):
                     ht=hl.read_table(args.ht_in_path),
                     output_ht_path=args.ht_out_path,
                     overwrite_temp=args.overwrite_temp,
+                    model_train_fold=args.model_train_fold,
+                    use_release=args.use_release,
+                    temp_label=args.temp_label,
                     freeze=args.freeze,
                 )
 
@@ -138,6 +156,15 @@ if __name__ == "__main__":
     parser.add_argument(
         "--slack-channel",
         help="Send message to Slack channel/user.",
+    )
+    parser.add_argument(
+        "--do-k-fold-training",
+        help="""
+        Generate (or prepare generation of) k-fold MPC models trained on training sets from respective folds.
+        
+        Otherwise, one model is generated, trained on all training transcripts.
+        """,
+        action="store_true",
     )
 
     # Create subparsers for each step
@@ -167,6 +194,16 @@ if __name__ == "__main__":
         """,
     )
     run_glm.add_argument(
+        "--use-model-formula",
+        help="Use specified model formula in regression.",
+        action="store_true",
+    )
+    run_glm.add_argument(
+        "--model-formula",
+        help="R-style model formula to use in regression. Required if --use-model-formula is set.",
+        default="pop_v_path ~ oe + misbad + oe:misbad + polyphen + oe:polyphen",
+    )
+    run_glm.add_argument(
         "--variables",
         help="Comma separated string of variables to include in all logistic regression.",
         default="oe,misbad,polyphen",
@@ -187,6 +224,26 @@ if __name__ == "__main__":
     annotate_hts = subparsers.add_parser(
         "annotate-hts", help="Annotate specified dataset with MPC."
     )
+    annotate_hts.add_argument(
+        "--model-train-fold",
+        help="Fold number in training set used to generate MPC model.",
+    )
+    annotate_hts.add_argument(
+        "--use-release",
+        help="""
+        Use scores in MPC release HT to annotate.
+
+        Otherwise, scores will be directly computed using specified MPC model.
+        """,
+        action="store_true",
+    )
+    annotate_hts.add_argument(
+        "--temp-label",
+        help="""
+        Suffix to add to temporary data paths to avoid conflicting names for different models.
+        """,
+    )
+    # temp_label="_clinvar",
     annotate_hts.add_argument(
         "--clinvar", help="Calculate MPC for ClinVar variants", action="store_true"
     )
@@ -213,7 +270,6 @@ if __name__ == "__main__":
         "--ht-out-path",
         help="Output path for hail Table after adding MPC annotation. Required if --specify-ht is set.",
     )
-    # TODO: Review if all these args are needed
 
     args = parser.parse_args()
 
--- a/rmc/pipeline/calculate_mpc.py
+++ b/rmc/pipeline/calculate_mpc.py
@@ -66,59 +66,16 @@ def main(args):
         if args.command == "annotate-hts":
             hl.init(log="/annotate_hts.log", tmp_dir=temp_dir)
             mpc_bucket_path = f"{MPC_PREFIX}/{CURRENT_GNOMAD_VERSION}/{args.freeze}"
-            if args.clinvar:
-                from rmc.resources.reference_data import clinvar_plp_mis_haplo
-
-                annotate_mpc(
-                    ht=clinvar_plp_mis_haplo.ht(),
-                    output_ht_path=f"{mpc_bucket_path}/clinvar_mpc_annot.ht",
-                    temp_label="_clinvar",
-                    model_train_fold=args.model_train_fold,
-                    use_release=args.use_release,
-                    overwrite_temp=args.overwrite_temp,
-                    freeze=args.freeze,
-                )
-                # TODO: Add support for annotating with models from specific folds (all folds?)
+            # TODO: Add support for annotating with models from specific folds (all folds?)
 
             if args.dd:
                 from rmc.resources.reference_data import ndd_de_novo
 
                 dd_ht = ndd_de_novo.ht()
-                dd_ht = dd_ht.explode("sample_set")
-                case_ht = dd_ht.filter(dd_ht.sample_set != "control")
                 annotate_mpc(
-                    ht=case_ht,
-                    output_ht_path=f"{mpc_bucket_path}/dd_case_mpc_annot.ht",
-                    temp_label="_dd_case",
-                    model_train_fold=args.model_train_fold,
-                    use_release=args.use_release,
-                    overwrite_temp=args.overwrite_temp,
-                    freeze=args.freeze,
-                )
-                control_ht = dd_ht.filter(dd_ht.sample_set == "control")
-                shared_variants = case_ht.join(control_ht, how="inner")
-                logger.info(
-                    "%i variants overlapped between cases and controls",
-                    shared_variants.count(),
-                )
-                annotate_mpc(
-                    ht=control_ht,
-                    output_ht_path=f"{mpc_bucket_path}/dd_control_mpc_annot.ht",
-                    temp_label="_dd_control",
-                    model_train_fold=args.model_train_fold,
-                    use_release=args.use_release,
-                    overwrite_temp=args.overwrite_temp,
-                    freeze=args.freeze,
-                )
-
-            if args.gnomad_exomes:
-                from gnomad.resources.grch37.gnomad import public_release
-
-                ht = public_release("exomes").ht()
-                annotate_mpc(
-                    ht=ht,
-                    output_ht_path=f"{mpc_bucket_path}/gnomAD_mpc_annot.ht",
-                    temp_label="_gnomad_exome",
+                    ht=dd_ht,
+                    output_ht_path=f"{mpc_bucket_path}/dd_mpc_annot.ht",
+                    temp_label="_dd",
                     model_train_fold=args.model_train_fold,
                     use_release=args.use_release,
                     overwrite_temp=args.overwrite_temp,
--- a/rmc/pipeline/calculate_mpc.py
+++ b/rmc/pipeline/calculate_mpc.py
@@ -199,18 +199,9 @@ if __name__ == "__main__":
         Suffix to add to temporary data paths to avoid conflicting names for different models.
         """,
     )
-    # temp_label="_clinvar",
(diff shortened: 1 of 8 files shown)


rmc/resources/rmc.py
```diff
@@ -624,19 +613,6 @@ def gnomad_fitted_score_path(
...
-            path=f"{MPC_PREFIX}/{CURRENT_GNOMAD_VERSION}/{freeze}/mpc_dedup.ht"
-        )
-        for freeze in FREEZES
-    },
-)
-"""
-Table containing missense variants in canonical transcripts annotated with MPC.
 
 This Table contains only one row per each unique locus/alleles combination.
 """
```
Ruchit10, line 618: There seems to be a bug in line 623 with variable `oe_bin_counts_tsv` where `{CURRENT_FREEZE}` is written twice and really should be `{CURRENT_GNOMAD_VERSION}` ?

rmc/utils/mpc.py
```diff
@@ -194,20 +199,19 @@ def prepare_pop_path_ht(
     :param bool overwrite_temp: Whether to overwrite intermediate temporary data if it already exists.
         If False, will read existing intermediate temporary data rather than overwriting.
         Default is False.
-    :param bool overwrite_output: Whether to entirely overwrite final output data if it already exists.
-        If False, will read and modify existing output data by adding or modifying columns rather than overwriting entirely.
-        If True, will clear existing output data and write new output data.
-        The output Tables are the OE-annotated context Tables with duplicated or deduplicated sections
-        and the population/pathogenic variant Table.
-        Default is True.
+    :param do_k_fold_training: Whether to generate k-fold Tables with the training transcripts.
```
Ruchit10, line 202:
```suggestion
    :param bool do_k_fold_training: Whether to generate k-fold Tables with the training transcripts.
```

rmc/utils/mpc.py
```diff
@@ -517,44 +504,191 @@ def _get_min_aic_glm(
...
+        If True, only one model will be built using the formula specified in `model_formula`.
+        If False, all possible combinations of variables in `variables` and `additional_variables`
+        will be used to build models. The single best model based on AIC will be kept.
+        Default is False.
+    :param str model_formula: Model formula to use in regression.
+        Default is 'pop_v_path ~ oe + misbad + oe:misbad + polyphen + oe:polyphen'.
+        This was the formula selected in ExAC.
+    :param List[str] variables: Variables to include in all regressions (single, joint).
+        Default is ['oe', 'misbad', 'polyphen'].
+    :param List[str] additional_variables: Additional variables to include in single variable regressions only.
```
Ruchit10, line 553:
```suggestion
    :param List[str] additional_variables: Additional variables to include in all regressions (single, joint).
```

rmc/utils/mpc.py
```diff
@@ -517,44 +504,191 @@ def _get_min_aic_glm(
...
+
+    :param bool use_model_formula: Whether to use a specified model formula.
+        If True, only one model will be built using the formula specified in `model_formula`.
+        If False, all possible combinations of variables in `variables` and `additional_variables`
+        will be used to build models. The single best model based on AIC will be kept.
+        Default is False.
+    :param str model_formula: Model formula to use in regression.
+        Default is 'pop_v_path ~ oe + misbad + oe:misbad + polyphen + oe:polyphen'.
+        This was the formula selected in ExAC.
+    :param List[str] variables: Variables to include in all regressions (single, joint).
```
Ruchit10, line 551:
```suggestion
    :param List[str] variables: Primary variables to include in all regressions (single, joint).
```

rmc/utils/mpc.py
```diff
@@ -517,44 +504,191 @@ def _get_min_aic_glm(
...
+    """
+    if not use_model_formula and do_k_fold_training:
+        raise DataException(
+            "Training separate models with k folds may result in different optimums!"
+        )
+
+    def _generate_regression_model(
+        temp_label: str,
+        fold: int = None,
+    ):
```
Ruchit10, line 573:
```suggestion
    ) -> None:
```

rmc/utils/mpc.py
```diff
@@ -669,158 +857,177 @@ def aggregate_gnomad_fitted_scores(
...
+        ht = ht.annotate(mpc=scores_ht[ht.key].mpc)
+        ht.write(output_ht_path, overwrite=True)
 
 
-def annotate_mpc(
-    ht: hl.Table,
-    output_path: str,
-    overwrite: bool = True,
-    add_transcript_annotation: bool = True,
+def create_mpc_release_ht(
```
ch-kr, line 1013: maybe we should remove this function entirely after your reorganization of `annotate_mpc`? this part
```
    annotate_mpc(
        ht=context_with_oe.versions[freeze].ht().select(),
        output_ht_path=mpc_release.versions[freeze].path,
        temp_label="_release",
        use_release=False,
        overwrite_temp=overwrite_temp,
        freeze=freeze,
    )
```
could get called in `calculate_mpc.py`

rmc/pipeline/calculate_mpc.py
```diff
@@ -37,24 +37,26 @@ def main(args):
...
+                overwrite_temp=args.overwrite_temp,
+                do_k_fold_training=args.do_k_fold_training,
                 freeze=args.freeze,
             )
 
-        if args.command == "calculate-mpc":
-            hl.init(log="/calculate_mpc_release.log", tmp_dir=temp_dir)
+        if args.command == "create-mpc-release":
+            hl.init(log="/create_mpc_release.log", tmp_dir=temp_dir)
             create_mpc_release_ht(
```
ch-kr, line 58: maybe we should remove this function (see comment in `mpc.py`)

rmc/pipeline/calculate_mpc.py
```diff
@@ -91,8 +100,11 @@ def main(args):
                 )
                 annotate_mpc(
```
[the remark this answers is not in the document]
ch-kr: also: the gnomAD exomes and ClinVar args were added for our convenience, but we could remove those sections as well

rmc/utils/mpc.py
ch-kr, line 1: `isort` reorganizes the imports in this file

rmc/utils/mpc.py
```diff
@@ -517,44 +504,191 @@ def _get_min_aic_glm(
...
 
-    logger.info(
-        "Annotating gnomAD variants with fitted score and writing to gnomad_fitted_score path..."
-    )
-    ht = ht.filter(ht.pop_v_path == 1)
-    ht = calculate_fitted_scores(ht=ht, freeze=freeze)
-    ht.write(gnomad_fitted_score_path(freeze=freeze), overwrite=overwrite)
+
+def run_regressions(
+    use_model_formula: bool = False,
```
ch-kr, line 511: is this just to make it easier for our team tor run just one formula (if we choose one specific formula to go with for v2)?

rmc/resources/rmc.py
ch-kr: good catch!

ch-kr (changes requested):
whew, this was a big one -- glad we have two people reviewing!

rmc/utils/mpc.py
Ruchit10: Ah that's more sensible!

rmc/utils/mpc.py
Ruchit10: Just for consistency

rmc/utils/mpc.py
```diff
@@ -517,44 +504,191 @@ def _get_min_aic_glm(
...
-    :return: Table filtered to defined MPC model annotations and annotated with fitted score.
+    :param int freeze: RMC data freeze number. Default is CURRENT_FREEZE.
+    :return: Table annotated with MPC model features and fitted scores, filtered to variants with defined annotations.
     """
     assert interaction_char in {"*", ":"}, "interaction_char must be one of '*' or ':'!"
 
     logger.info("Extracting MPC model relationships from pickle...")
-    model_path = mpc_model_pkl_path(freeze)
-    with hl.hadoop_open(model_path, "rb") as p:
+    with hl.hadoop_open(mpc_model_pkl_path(fold=fold, freeze=freeze), "rb") as p:
```
Ruchit10, line 691: Should there be a check here for the saved model pickle and raise exception if not? Possibly because this function is also called in `annotate_mpc` which could be run independently?

rmc/utils/mpc.py
lw453: Good catch, thanks!

rmc/utils/mpc.py
lw453: Yes, this is cited in our [`Cosmetic fixes`](https://github.com/broadinstitute/rmc_production/issues/108) issue:
```
Remove param types in docstrings
```
So I haven't really been bothering to make these consistent right now

rmc/resources/reference_data.py
```diff
@@ -66,41 +66,46 @@
...
     if fold is not None and fold not in range(1, FOLD_K + 1):
         raise DataException(
             f"Fold number must be an integer between 1 and {FOLD_K} inclusive!"
         )
 
-    train_type = "val" if is_val else "train"
+    transcript_type = "test" if is_test else ("val" if is_val else "train")
     fold_name = f"_fold{fold}" if fold is not None else ""
-    return f"{RESOURCE_BUILD_PREFIX}/{train_type}_transcripts{fold_name}.he"
+    return f"{RESOURCE_BUILD_PREFIX}/{transcript_type}_transcripts{fold_name}.he"
```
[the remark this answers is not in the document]
lw453: If `fold` is None, `"{str(fold)}"` will give `"None"`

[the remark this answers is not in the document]
ch-kr: oh wow it does, I thought I tested this locally

rmc/resources/reference_data.py
```diff
@@ -66,41 +66,46 @@
...
+    :param is_test: Whether to return test transcripts.
+        If False, training transcripts will be returned. If True, test transcripts will be returned.
+        Default is False.
     :param fold: Fold number in training set to select transcripts from.
         If None, all training transcripts will be returned.
-        If not None, only validation or training transcripts from the specified fold
-            will be returned.
-        Default is None.
+        If not None, only validation or training transcripts from the specified fold will be returned.
+        Default is None. NOTE that `is_test` must be False if `fold` is not None.
```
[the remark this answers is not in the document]
lw453: Good idea!

rmc/utils/missense_badness.py
```diff
@@ -299,170 +277,122 @@ def calculate_misbad(
...
+            )
+            oe_hts[oe_direction] = aggregate_aa_and_filter_oe(
+                ht, keep_high_oe=keep_high_oe
+            )
+            oe_hts[oe_direction] = oe_hts[oe_direction].transmute(
+                **{
+                    f"{oe_direction}_obs": oe_hts[oe_direction].obs,
+                    f"{oe_direction}_pos": oe_hts[oe_direction].possible,
+                }
+            )
```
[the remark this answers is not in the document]
lw453: Yes! Good catch

rmc/pipeline/regional_constraint.py
```diff
@@ -664,6 +665,14 @@ def main(args):
     parser.add_argument(
         "--overwrite", help="Overwrite existing data.", action="store_true"
```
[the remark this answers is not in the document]
lw453: Good catches

rmc/utils/mpc.py
```diff
@@ -564,77 +698,127 @@ def calculate_fitted_scores(
...
             import_grantham()
         grantham_ht = grantham.ht()
-        ht = ht.annotate(grantham=grantham_ht[ht.ref, ht.alt].score)
-        annotations.append("grantham")
+        scores_ht = scores_ht.annotate(
+            grantham=grantham_ht[scores_ht.ref, scores_ht.alt].score
+        )
+        annots.add("grantham")
+
+    if annots != variables:
```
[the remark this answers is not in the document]
lw453: I see what you mean. The point of this check was to make sure the resulting table contains all the annotations we want, so I added another check at the beginning.

rmc/utils/mpc.py
lw453: Yes

rmc/utils/mpc.py
```diff
@@ -564,77 +698,127 @@ def calculate_fitted_scores(
...
+    variables = variable_dict.keys()
+    interactions_dict = {k: v for k, v in mpc_rel_vars.items() if interaction_char in k}
 
-    if "oe" in variables:
-        annotations.append("oe")
+    # Select fields from context HT containing annotations to extract
+    # NOTE: `ref` and `alt` fields are also extracted for annotation of missense badness
+    context_ht_annots = {"oe", "polyphen"}
+    annots = context_ht_annots.intersection(variables)
+    logger.info("Annotating HT with MPC variables from context HT (%s)...", annots)
```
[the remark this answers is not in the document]
lw453: It's possible it may be empty if the MPC model doesn't include any annotations we want to pull from the context table. But also check out the new code and lmk what you think!

rmc/utils/mpc.py
lw453: Yes, great point

rmc/utils/mpc.py
```diff
@@ -297,80 +294,97 @@ def prepare_pop_path_ht(
...
 
-def run_regressions(
-    variables: List[str] = ["oe", "misbad", "polyphen"],
-    additional_variables: List[str] = ["blosum", "grantham"],
-    overwrite: bool = True,
-    freeze: int = CURRENT_FREEZE,
-) -> None:
-    """
-    Run single variable and joint regressions and pick best model.
+    def _add_misbad_and_filter_transcripts(
```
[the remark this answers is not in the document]
lw453: I placed it here because if we do it this way, all the folds can use the same base table to filter to their training transcripts. Otherwise we would have to redo all the filtering steps for each fold

rmc/pipeline/calculate_mpc.py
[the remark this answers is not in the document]
lw453: Done!

rmc/pipeline/calculate_mpc.py
Ruchit10, line 1: Based on the changes are `--clinvar` and `--gnomad-exomes` sub-arguments defined in lines 203 and 211 still required or are obsolete at this point?

Ruchit10 (approved):
Looks good. Just mentioned a comment about a couple extra flags in the MPC pipeline script that might be redundant

rmc/pipeline/calculate_mpc.py
lw453: Thanks for catching! Have removed now

rmc/utils/mpc.py
```diff
@@ -702,18 +700,25 @@ def calculate_fitted_scores(
     variables = variable_dict.keys()
     interactions_dict = {k: v for k, v in mpc_rel_vars.items() if interaction_char in k}
 
+    context_ht_annots = {"oe", "polyphen"}.intersection(variables)
```
ch-kr, line 703: I like adding a check here, but this doesn't check that the values exist on the context HT, right?

Maybe this check could get combined with the one below to look something like: 
```
context_ht_annots = {"oe", "polyphen"}.intersection(variables).union(helper_annots)
context_ht = context_with_oe.versions[freeze].ht()
annots_check = {field for field in context_ht.row} - context_ht_annots
if len(annots_check) != 0:
    raise DataException(f"{annots_check} missing from context HT!")
```
I also wonder if we should pass in some of these hardcoded sets (e.g., `context_ht_annots`, `helper_annots`) as args with default values?

ch-kr (commented):
just one more small comment

ch-kr (approved):
LGTM!

PR #294, opened 2023-07-20
Browser release updates

ch-kr:
This PR contains code updates required to reformat RMC results for gnomAD browser release.

Major updates:
`rmc/utils/generic.py`
- Added two new functions to get reference amino acid information (3 letter code + number) from context HT

`rmc/utils/constraint.py`
- Added functions to annotate RMC results HT with amino acids, then check for and fix any missing amino acid information
- Updated `reformat_annotations_for_release` to reformat RMC HT to new desired schema
- Added a function to annotate the globals of the RMC HT for release

Minor changes:
`rmc/pipeline/regional_constraint.py`
- Updated code to annotate CDS start and stop positions as transcript start and stops (rather than browser start/stops)
- Added call to `reformat_annotations_for_release`

### DIFF
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -303,7 +303,9 @@ def main(args):
                     logger.info(
                         "Adding section annotation before searching for first break..."
                     )
-                    # Add transcript start and stop positions from browser HT
+                    # Add transcript start and stop positions from CDS HT
+                    # NOTE: For RMC freezes 1-7, the transcript start and ends were from the
+                    # browser HT (`gene_model`) rather than the CDS HT!
                     transcript_ht = transcript_cds.ht().key_by()
                     transcript_ht = transcript_ht.annotate(
                         start=transcript_ht.interval.start.position,
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -40,6 +40,7 @@ from rmc.utils.constraint import (
     get_max_chisq_per_group,
     merge_rmc_hts,
     process_sections,
+    reformat_annotations_for_release,
 )
 from rmc.utils.generic import (
     filter_context_using_gnomad,
@@ -597,6 +598,9 @@ def main(args):
             create_context_with_oe(
                 freeze=args.freeze, overwrite_temp=args.overwrite_temp
             )
+
+            logger.info("Reformatting RMC results for browser release...")
+            reformat_annotations_for_release(args.freeze, args.overwrite_temp)
 
     finally:
         logger.info("Copying hail log to logging bucket...")
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -17,7 +17,7 @@ from rmc.resources.gnomad import (
     processed_exomes,
     prop_obs_coverage,
 )
-from rmc.resources.reference_data import filtered_context, gene_model
+from rmc.resources.reference_data import filtered_context, transcript_cds
 from rmc.resources.resource_utils import MISSENSE
 from rmc.resources.rmc import (
     CURRENT_FREEZE,
@@ -304,7 +304,11 @@ def main(args):
                         "Adding section annotation before searching for first break..."
                     )
                     # Add transcript start and stop positions from browser HT
-                    transcript_ht = gene_model.ht().select("start", "stop")
+                    transcript_ht = transcript_cds.ht().key_by()
+                    transcript_ht = transcript_ht.annotate(
+                        start=transcript_ht.interval.start.position,
+                        stop=transcript_ht.interval.end.position,
+                    )
                     ht = ht.annotate(**transcript_ht[ht.transcript])
                     ht = ht.annotate(
                         section=hl.format("%s_%s_%s", ht.transcript, ht.start, ht.stop)
--- a/rmc/utils/constraint.py
+++ b/rmc/utils/constraint.py
@@ -1334,7 +1334,8 @@ def annot_rmc_with_aa(ht: hl.Table, overwrite_temp: bool):
         overwrite_temp=overwrite_temp, keep_transcripts=rmc_transcripts
     )
     context_ht = get_ref_aa(
-        ht=context_ht, overwrite_temp=overwrite_temp, collect_all_aa=False
+        ht=context_ht,
+        overwrite_temp=overwrite_temp,
     )
     ht = ht.annotate(
         start_aa=context_ht[ht.start_coordinate, ht.transcript].ref_aa,
@@ -1602,7 +1603,10 @@ def reformat_annotations_for_release(freeze: int, overwrite_temp: bool) -> None:
         stop_coordinate=ht.interval.end,
     )
 
+    # Annotate amino acids
     ht = annot_rmc_with_aa(ht, overwrite_temp)
+
+    # Add region struct
     ht = ht.annotate(
         regions=hl.struct(
             start_coordinate=ht.start_coordinate,
--- a/rmc/utils/constraint.py
+++ b/rmc/utils/constraint.py
@@ -18,7 +18,12 @@ from rmc.resources.basics import (
     TEMP_PATH_WITH_SLOW_DEL,
 )
 from rmc.resources.gnomad import constraint_ht, filtered_exomes
-from rmc.resources.reference_data import clinvar_plp_mis_haplo, gene_model, ndd_de_novo
+from rmc.resources.reference_data import (
+    clinvar_plp_mis_haplo,
+    gene_model,
+    ndd_de_novo,
+    transcript_cds,
+)
 from rmc.resources.resource_utils import MISSENSE
 from rmc.resources.rmc import (
     CONSTRAINT_ANNOTATIONS,
@@ -40,9 +45,11 @@ from rmc.resources.rmc import (
     single_search_round_ht_path,
 )
 from rmc.utils.generic import (
+    get_aa_from_context,
     get_annotations_from_context_ht_vep,
     get_constraint_transcripts,
     get_coverage_correction_expr,
+    get_ref_aa,
     import_clinvar,
     import_de_novo_variants,
     keep_criteria,
@@ -1311,7 +1318,247 @@ def create_context_with_oe(
     logger.info("Output OE-annotated dedup context HT fields: %s", set(ht.row))
 
 
-def reformat_annotations_for_release(freeze: int, overwrite: bool) -> None:
+def annot_rmc_with_aa(ht: hl.Table, overwrite_temp: bool):
+    """
+    Annotate RMC results HT with amino acid information.
+
+    :param ht: Input RMC results HT.
+    :param overwrite_temp: Whether to overwrite temporary data.
+        If False, will read existing temp data rather than overwriting.
+        If True, will overwrite temp data.
+    :return: RMC results HT annotated with amino acid information.
+    """
+    logger.info("Getting amino acid information from context HT...")
+    rmc_transcripts = ht.aggregate(hl.agg.collect_as_set(ht.transcript))
+    context_ht = get_aa_from_context(
+        overwrite_temp=overwrite_temp, keep_transcripts=rmc_transcripts
+    )
+    context_ht = get_ref_aa(
+        ht=context_ht, overwrite_temp=overwrite_temp, collect_all_aa=False
+    )
+    ht = ht.annotate(
+        start_aa=context_ht[ht.start_coordinate, ht.transcript].ref_aa,
+        stop_aa=context_ht[ht.stop_coordinate, ht.transcript].ref_aa,
+    )
+    ht = ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc_results_aa_annot.ht",
+        _read_if_exists=not overwrite_temp,
+        overwrite=overwrite_temp,
+    )
+    return check_and_fix_missing_aa(ht, context_ht, overwrite_temp)
+
+
+def join_and_fix_aa(ht: hl.Table, fix_ht: hl.Table) -> hl.Table:
+    """
+    Add start and stop amino acid information from fix_ht to ht.
+
+    :param ht: Input HT. Should be RMC results HT annotated with amino acid information.
+    :param fix_ht: HT start and stop amino acids only for RMC regions that were
+        previously missing amino acid information.
+    :return: RMC results HT annotated with amino acid information from fix HT.
+    """
+    ht = ht.annotate(
+        start_aa=hl.if_else(
+            hl.is_missing(ht.start_aa) & ht.transcript_start,
+            fix_ht[ht.interval, ht.transcript].start_aa,
+            ht.start_aa,
+        ),
+        stop_aa=hl.if_else(
+            hl.is_missing(ht.stop_aa) & ht.transcript_stop,
+            fix_ht[ht.interval, ht.transcript].stop_aa,
+            ht.stop_aa,
+        ),
+    )
+    return ht
+
+
+def fix_transcript_start_stops(ht: hl.Table, overwrite_temp: bool) -> hl.Table:
+    """
+    Fix missing start or stop amino acid at transcript start or stop coordinates.
+
+    :param ht: Input HT. Should be RMC results HT annotated with amino acid information.
+    :param overwrite_temp: Whether to overwrite temporary data.
+        If False, will read existing temp data rather than overwriting.
+        If True, will overwrite temp data.
+    :return: HT containing start and stop amino acids only for RMC regions that were
+        previously missing amino acid information.
+    """
+    miss_start_stop_ht = ht.filter(
+        (hl.is_missing(ht.start_aa) & ht.transcript_start)
+        | (hl.is_missing(ht.stop_aa) & ht.transcript_stop)
+    )
+    miss_start_stop_ht = miss_start_stop_ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/transcript_start_stop_missing_aa.ht",
+        _read_if_exists=not overwrite_temp,
+        overwrite=overwrite_temp,
+    )
+    miss_start_stop_transcripts = miss_start_stop_ht.aggregate(
+        hl.agg.collect_as_set(miss_start_stop_ht.transcript)
+    )
+    filt_context_ht = get_aa_from_context(
+        overwrite_temp=False, keep_transcripts=miss_start_stop_transcripts
+    )
+    filt_context_ht = filt_context_ht.filter(
+        hl.literal(miss_start_stop_transcripts).contains(filt_context_ht.transcript)
+    )
+    filt_context_ht = filt_context_ht.group_by("transcript").aggregate(
+        max_aa_num=hl.agg.max(filt_context_ht.protein_start)
+    )
+    filt_context_ht = filt_context_ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/transcript_start_stop_missing_max_aa_num.ht",
+        _read_if_exists=not overwrite_temp,
+        overwrite=overwrite_temp,
+    )
+    miss_start_stop_ht = miss_start_stop_ht.annotate(
+        **filt_context_ht[miss_start_stop_ht.transcript]
+    )
+    miss_start_stop_ht = miss_start_stop_ht.annotate(
+        start_aa=hl.if_else(
+            hl.is_missing(miss_start_stop_ht.start_aa),
+            hl.if_else(
+                # Set all missing + strand transcript starts to Met1
+                miss_start_stop_ht.strand == "+",
+                "Met1",
+                hl.format("Ter%s", miss_start_stop_ht.max_aa_num),
+            ),
+            miss_start_stop_ht.start_aa,
+        ),
+        stop_aa=hl.if_else(
+            hl.is_missing(miss_start_stop_ht.stop_aa),
+            hl.if_else(
+                # Set all missing - strand transcript stops to Met1
+                miss_start_stop_ht.strand == "-",
+                "Met1",
+                hl.format("Ter%s", miss_start_stop_ht.max_aa_num),
+            ),
+            miss_start_stop_ht.start_aa,
+        ),
+    )
+    return miss_start_stop_ht.select("start_aa", "stop_aa")
+
+
+def fix_region_start_stops(
+    ht: hl.Table, context_ht: hl.Table, overwrite_temp: bool
+) -> hl.Table:
+    """
+    Fix missing start or stop amino acid information for non-transcript starts and stops.
+
+    :param ht: Input HT. Should be RMC results HT annotated with amino acid information.
+    :param overwrite_temp: Whether to overwrite temporary data.
+        If False, will read existing temp data rather than overwriting.
+        If True, will overwrite temp data.
+    :return: HT containing start and stop amino acids only for RMC regions that were
+        previously missing amino acid information.
+    """
+    missing_ht = ht.filter(hl.is_missing(ht.start_aa) | hl.is_missing(ht.stop_aa))
+    missing_transcripts = missing_ht.aggregate(
+        hl.agg.collect_as_set(missing_ht.transcript)
+    )
+    filt_context_ht = context_ht.filter(
+        hl.literal(missing_transcripts).contains(context_ht.transcript)
+    )
+    pos_ht = constraint_prep.ht().select_globals().select()
+    pos_ht = pos_ht.filter(hl.literal(missing_transcripts).contains(pos_ht.transcript))
+    pos_ht = pos_ht.annotate(pos=pos_ht.locus.position)
+    pos_ht = pos_ht.key_by("locus")
+    pos_ht = pos_ht.group_by("transcript").aggregate(
+        positions=hl.agg.collect(pos_ht.pos)
+    )
+    pos_ht = pos_ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/pos_per_transcript.ht",
+        _read_if_exists=not overwrite_temp,
+        overwrite=overwrite_temp,
+    )
+    missing_ht = missing_ht.annotate(
+        # Take the first missense position that is >= start coordinate
+        closest_start_mis=hl.bind(
+            lambda x: x[hl.binary_search(x, missing_ht.start_coordinate.position)],
+            pos_ht[missing_ht.transcript].positions,
+        ),
+        # Take the first missense position that is <= stop coordinate
+        closest_stop_mis=hl.bind(
+            lambda x: x[hl.binary_search(x, missing_ht.stop_coordinate.position) - 1],
+            pos_ht[missing_ht.transcript].positions,
+        ),
+    )
+    # Checkpoint to make sure the binary search/join only runs once,
+    # especially since there is another join immediately afterwards
+    missing_ht = missing_ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/region_missing_start_stop_pos.ht",
+        _read_if_exists=not overwrite_temp,
+        overwrite=overwrite_temp,
+    )
+    missing_ht = missing_ht.annotate(
+        start_aa=hl.if_else(
+            hl.is_missing(missing_ht.start_aa),
+            filt_context_ht[missing_ht.closest_start_mis, missing_ht.transcript].ref_aa,
+            missing_ht.start_aa,
+        ),
+        stop_aa=hl.if_else(
+            hl.is_missing(missing_ht.stop_aa),
+            filt_context_ht[missing_ht.closest_stop_mis, missing_ht.transcript].ref_aa,
+            missing_ht.stop_aa,
+        ),
+    )
+    return missing_ht.select("start_aa", "stop_aa")
+
+
+def check_and_fix_missing_aa(
+    ht: hl.Table, context_ht: hl.Table, overwrite_temp: bool
+) -> hl.Table:
+    """
+    Check for and fix any missing amino acid information.
+
+    :param ht: Input Table. Should be RMC results HT annotated with amino acid information.
+    :param context_ht: VEP context HT filtered to keep only transcript ID, protein number, and amino acid information.
+    :param overwrite_temp: Whether to overwrite temporary data.
+        If False, will read existing temp data rather than overwriting.
+        If True, will overwrite temp data.
+    :return: RMC results HT annotated with amino acids for all region start/stops.
+    """
+    missing_ht = ht.filter(hl.is_missing(ht.start_aa) | hl.is_missing(ht.stop_aa))
+    missing_start_or_stop = missing_ht.count()
+    if missing_start_or_stop == 0:
+        return ht
+
+    logger.info(
+        "%i sites are missing either their stop or start AA!", missing_start_or_stop
+    )
+
+    # Get strand from browser HT
+    # TODO: recreate browser HT to include strand annotation
+    browser_ht = gene_model.ht().select("strand")
+    ht = ht.annotate(strand=gene_model[ht.transcript].strand)
+
+    # Get CDS start/stops from CDS HT
+    transcript_ht = transcript_cds.ht().key_by()
+    transcript_ht = transcript_ht.annotate(
+        cds_start=transcript_ht.interval.start,
+        cds_stop=transcript_ht.interval.end,
+    )
+    ht = ht.annotate(**transcript_ht[ht.transcript])
+    ht = ht.annotate(
+        transcript_start=ht.start_coordinate == ht.cds_start,
+        transcript_stop=ht.stop_coordinate == ht.cds_stop,
+    )
+    transcript_start_stop_fix_ht = fix_transcript_start_stops(ht, overwrite_temp)
+    ht = join_and_fix_aa(ht, transcript_start_stop_fix_ht)
+    ht = ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/RMC_results_transcript_start_stop_aa_fix.ht",
+        _read_if_exists=not overwrite_temp,
+        overwrite=overwrite_temp,
+    )
+    region_fix_ht = fix_region_start_stops(ht, context_ht, overwrite_temp)
+    ht = join_and_fix_aa(ht, region_fix_ht)
+    ht = ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/RMC_results_all_aa_fix.ht",
+        _read_if_exists=not overwrite_temp,
+        overwrite=overwrite_temp,
+    )
+    return ht
+
+
+def reformat_annotations_for_release(freeze: int, overwrite_temp: bool) -> None:
     """
     Reformat annotations in input HT for release.
 
@@ -1342,10 +1589,20 @@ def reformat_annotations_for_release(freeze: int, overwrite: bool) -> None:
     ----------------------------------------
 
     :param freeze: RMC data freeze number.
-    :param overwrite: Whether to overwrite existing data.
+    :param overwrite_temp: Whether to overwrite temporary data.
+        If False, will read existing temp data rather than overwriting.
+        If True, will overwrite temp data.
     :return: None; writes Table with desired schema to resource path.
     """
     ht = rmc_results.versions[freeze].ht()
+
+    # Annotate start and stop coordinates per region
+    ht = ht.annotate(
+        start_coordinate=ht.interval.start,
+        stop_coordinate=ht.interval.end,
+    )
+
+    ht = annot_rmc_with_aa(ht, overwrite_temp)
     ht = ht.annotate(
         regions=hl.struct(
             start_coordinate=ht.start_coordinate,
@@ -1362,7 +1619,7 @@ def reformat_annotations_for_release(freeze: int, overwrite: bool) -> None:
     ht = ht.group_by(transcript_id=ht.transcript).aggregate(
         regions=hl.agg.collect(ht.regions)
     )
-    ht.write(rmc_browser.versions[freeze].path, overwrite=overwrite)
+    ht.write(rmc_browser.versions[freeze].path, overwrite=True)
 
 
 def check_loci_existence(ht1: hl.Table, ht2: hl.Table, annot_str: str) -> hl.Table:
--- a/rmc/utils/constraint.py
+++ b/rmc/utils/constraint.py
@@ -1559,6 +1559,40 @@ def check_and_fix_missing_aa(
     return ht
 
 
+def annotate_transcript_globals(ht: hl.Table) -> hl.Table:
+    """
+    Annotate HT globals with transcript information.
+
+    Function is used when reformatting RMC results for browser release.
+    Annotates:
+        - transcripts not searched for RMC
+        - transcripts without evidence of RMC
+        - outlier transcripts
+    :param HT: Input Table. Should be RMC results HT.
+    :return: RMC results HT with updated globals annotations.
+    """
+    # Get transcripts with evidence of RMC
+    rmc_transcripts = hl.literal(ht.aggregate(hl.agg.collect_as_set(ht.transcript)))
+
+    # Get all QC pass transcripts and outlier transcripts
+    qc_pass_transcripts = get_constraint_transcripts(outlier=False)
+    outlier_transcripts = get_constraint_transcripts(outlier=True)
+
+    # Get all transcripts displayed on browser
+    transcript_ht = gene_model.ht()
+    all_transcripts = transcript_ht.aggregate(
+        hl.agg.collect_as_set(transcript_ht.transcript)
+    )
+    transcripts_no_rmc = qc_pass_transcripts.difference(rmc_transcripts)
+    transcripts_not_searched = all_transcripts.difference(qc_pass_transcripts)
+    ht = ht.select_globals()
+    return ht.annotate_globals(
+        transcripts_not_searched=transcripts_not_searched,
+        transcripts_no_rmc=transcripts_no_rmc,
+        outlier_transcripts=outlier_transcripts,
+    )
+
+
 def reformat_annotations_for_release(freeze: int, overwrite_temp: bool) -> None:
     """
     Reformat annotations in input HT for release.
@@ -1623,6 +1657,9 @@ def reformat_annotations_for_release(freeze: int, overwrite_temp: bool) -> None:
     ht = ht.group_by(transcript_id=ht.transcript).aggregate(
         regions=hl.agg.collect(ht.regions)
     )
+
+    # Annotate globals and write
+    ht = annotate_transcript_globals(ht)
     ht.write(rmc_browser.versions[freeze].path, overwrite=True)
 
 
--- a/rmc/utils/constraint.py
+++ b/rmc/utils/constraint.py
@@ -32,6 +32,7 @@ from rmc.resources.rmc import (
     context_with_oe_dedup,
     no_breaks_he_path,
     oe_bin_counts_tsv,
+    rmc_browser,
     rmc_results,
     simul_search_bucket_path,
     simul_search_round_bucket_path,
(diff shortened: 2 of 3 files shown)


rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+                "Met1",
+                hl.format("Ter%s", miss_start_stop_ht.max_aa_num),
+            ),
+            miss_start_stop_ht.start_aa,
+        ),
+    )
+    return miss_start_stop_ht.select("start_aa", "stop_aa")
+
+
+def fix_region_start_stops(
```
lw453, line 1441: Hmm as I understand it this introduces some imprecision (albeit only for a small number of transcripts) to the breakpoint coordinates that we report. An alternative is to calculate the exact amino acid number by counting up the coding sequence lengths up to the missing AA (and just not report the actual AA identity if not available). Or we could just not report the AAs in the tooltips for these regions entirely.

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+    Add start and stop amino acid information from fix_ht to ht.
+
+    :param ht: Input HT. Should be RMC results HT annotated with amino acid information.
+    :param fix_ht: HT start and stop amino acids only for RMC regions that were
+        previously missing amino acid information.
+    :return: RMC results HT annotated with amino acid information from fix HT.
+    """
+    ht = ht.annotate(
+        start_aa=hl.if_else(
+            hl.is_missing(ht.start_aa) & ht.transcript_start,
```
lw453, line 1363: Doesn't this only work when you call it after `fix_transcript_start_stops` and not `fix_region_start_stops`?

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+        If False, will read existing temp data rather than overwriting.
+        If True, will overwrite temp data.
+    :return: HT containing start and stop amino acids only for RMC regions that were
+        previously missing amino acid information.
+    """
+    missing_ht = ht.filter(hl.is_missing(ht.start_aa) | hl.is_missing(ht.stop_aa))
+    missing_transcripts = missing_ht.aggregate(
+        hl.agg.collect_as_set(missing_ht.transcript)
+    )
+    filt_context_ht = context_ht.filter(
```
lw453, line 1458: This table isn't used

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+        stop_aa=hl.if_else(
+            hl.is_missing(ht.stop_aa) & ht.transcript_stop,
+            fix_ht[ht.interval, ht.transcript].stop_aa,
+            ht.stop_aa,
+        ),
+    )
+    return ht
+
+
+def fix_transcript_start_stops(ht: hl.Table, overwrite_temp: bool) -> hl.Table:
```
lw453, line 1376:
```suggestion
def fix_transcript_start_stop_aas(ht: hl.Table, overwrite_temp: bool) -> hl.Table:
```

rmc/utils/constraint.py
lw453, line 1441:
```suggestion
def fix_region_start_stop_aas(
```

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
-        'transcript_id': str
-        'regions': array<struct {
-            'start': int32
-            'stop': int32
-            'observed_missense': int32
-            'expected_missense': float64
-            'chisq': float64
-        }>
-
+        'interval': interval<locus<GRCh37>>
```
lw453, line 1610:
```suggestion
```

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+            start_aa: str,
+            stop_aa: str,
+            obs: int64,
+            exp: float64,
+            oe: float64,
+            chisq: float64
+        }
     ----------------------------------------
-    Key: ['transcript_id']
+    Key: ['interval', 'transcript']
```
lw453, line 1623:
```suggestion
    Key: ['transcript']
```

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+            stop_aa=ht.stop_aa,
+            obs=ht.section_obs,
+            exp=ht.section_exp,
+            oe=ht.section_oe,
             chisq=ht.section_chisq,
         )
     )
     # Group Table by transcript
-    return ht.group_by(transcript_id=ht.transcript).aggregate(
+    ht = ht.group_by(transcript_id=ht.transcript).aggregate(
```
lw453, line 1657:
```suggestion
    ht = ht.group_by("transcript").aggregate(
```

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+        start_coordinate=ht.interval.start,
+        stop_coordinate=ht.interval.end,
+    )
+
+    # Annotate amino acids
+    ht = annot_rmc_with_aa(ht, overwrite_temp)
+
+    # Add region struct
     ht = ht.annotate(
         regions=hl.struct(
```
lw453, line 1645:
```suggestion
        region=hl.struct(
```

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+            obs=ht.section_obs,
+            exp=ht.section_exp,
+            oe=ht.section_oe,
             chisq=ht.section_chisq,
         )
     )
     # Group Table by transcript
-    return ht.group_by(transcript_id=ht.transcript).aggregate(
+    ht = ht.group_by(transcript_id=ht.transcript).aggregate(
         regions=hl.agg.collect(ht.regions)
```
lw453, line 1658:
```suggestion
        regions=hl.agg.collect(ht.region)
```

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+    ht = join_and_fix_aa(ht, region_fix_ht)
+    ht = ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/RMC_results_all_aa_fix.ht",
+        _read_if_exists=not overwrite_temp,
+        overwrite=overwrite_temp,
+    )
+    return ht
+
+
+def annotate_transcript_globals(ht: hl.Table) -> hl.Table:
```
lw453, line 1562:
```suggestion
def add_globals_rmc_browser(ht: hl.Table) -> hl.Table:
```
Suggestion for function name

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+    transcripts_not_searched = all_transcripts.difference(qc_pass_transcripts)
+    ht = ht.select_globals()
+    return ht.annotate_globals(
+        transcripts_not_searched=transcripts_not_searched,
+        transcripts_no_rmc=transcripts_no_rmc,
+        outlier_transcripts=outlier_transcripts,
+    )
+
+
+def reformat_annotations_for_release(freeze: int, overwrite_temp: bool) -> None:
```
lw453, line 1596:
```suggestion
def format_rmc_browser(freeze: int, overwrite_temp: bool) -> None:
```
Suggestion for function name

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/RMC_results_all_aa_fix.ht",
+        _read_if_exists=not overwrite_temp,
+        overwrite=overwrite_temp,
+    )
+    return ht
+
+
+def annotate_transcript_globals(ht: hl.Table) -> hl.Table:
+    """
+    Annotate HT globals with transcript information.
```
lw453, line 1564:
```suggestion
    Annotate HT globals with RMC transcript information.
```

lw453 (changes requested):

rmc/utils/generic.py
```diff
@@ -220,6 +220,122 @@ def process_context_ht(
...
+    ht = ht.filter(
+        ~hl.literal(CSQ_NON_CODING).contains(
+            ht.transcript_consequences.most_severe_consequence
+        )
+    )
+    # Unnest annotations from context HT
+    ht = ht.transmute(
+        transcript=ht.transcript_consequences.trancript_id,
+        protein_start=ht.transcript_consequences.protein_start,
+        protein_end=ht.transcript_consequences.protein_end,
```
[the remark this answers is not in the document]
ch-kr: I took these right out of VEP ahha but agree!

rmc/utils/generic.py
```diff
@@ -220,6 +220,122 @@ def process_context_ht(
...
+        overwrite=overwrite_temp,
+    )
+    return ht
+
+
+def get_ref_aa(
+    ht: hl.Table,
+    overwrite_temp: bool,
+    extra_aa_map: Dict[str, str] = {"*": "Ter", "U": "Sec"},
+    aa_to_remove: str = "X",
```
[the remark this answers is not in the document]
ch-kr: sounds good!

rmc/utils/generic.py
```diff
@@ -220,6 +220,122 @@ def process_context_ht(
...
+    if extra_aa_map:
+        aa_map = aa_map.update(extra_aa_map)
+    ht = ht.annotate(ref_aa_1_letter=ht.amino_acids.split("/")[0])
+    ht = ht.annotate(
+        ref_aa=hl.literal(aa_map).get(ht.ref_aa_1_letter, ht.ref_aa_1_letter)
+    )
+    if aa_to_remove:
+        ht = ht.filter(ht.ref_aa != aa_to_remove)
+
+    # Double check that protein start always equals protein end
```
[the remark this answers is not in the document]
ch-kr: I think I did, which is why I added the "U"/"Sec" to the extra_aa_map and "X" to remove, but good idea!

rmc/utils/generic.py
```diff
@@ -220,6 +220,122 @@ def process_context_ht(
...
+            f"{protein_num_check} sites had different protein start and end values --"
+            " please double check!"
+        )
+    # Reformat reference AA to have both the 3 letter code and number
+    ht = ht.annotate(ref_aa=hl.format("%s%s", ht.ref_aa, ht.protein_start))
+
+    # Select fields and checkpoint
+    ht = ht.select("ref_aa", "transcript")
+    ht = ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/ref_amino_acids.ht",
```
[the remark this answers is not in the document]
ch-kr: not super needed; I kept it just to not clutter the top level `gs://gnomad-tmp-4day` bucket

rmc/utils/generic.py
```diff
@@ -220,6 +220,122 @@ def process_context_ht(
...
+        _read_if_exists=not overwrite_temp,
+        overwrite=overwrite_temp,
+    )
+
+    # Collect by key to collapse AA per locus
```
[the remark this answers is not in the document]
ch-kr: > It would shrink the table too.

that's a good point! I didn't add it to the top of the function so that I could (lazily) work with unnested annotations


> Also have you checked that when AA info is missing, it is missing for all rows for a locus-transcript combo? I.e. not just missing for one variant which we happen to pick up in this collapse. 

ah, I have not checked this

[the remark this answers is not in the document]
ch-kr: I might keep the collect by key where it is to make the ref aa check/protein number check more straightforward

rmc/utils/constraint.py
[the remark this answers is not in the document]
ch-kr: we could also adjust these edge cases to simply have the closest AA (same fix as `fix_region_start_stops`), which might be more consistent

rmc/utils/constraint.py
ch-kr: wait, it is, right? at the end

rmc/utils/constraint.py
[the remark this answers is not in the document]
ch-kr: ^I should note that this is actually what I did in the latest demo data for Riley

rmc/utils/generic.py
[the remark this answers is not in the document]
lw453: But `{TEMP_PATH_WITH_FAST_DEL}` already includes `/rmc/`

rmc/utils/constraint.py
ch-kr: good point, I'll add a note

rmc/utils/constraint.py
lw453: Oh you're right, it was marked as unused in my vscode and I must have not scrolled down far enough to check the whole function. Thanks for correcting me!

rmc/utils/constraint.py
ch-kr: we chatted about this briefly in slack, but when I spot checked region start/stops that were missing AAs, all of the positions I checked were in the intron. I think this is from how we define sections (e.g., https://github.com/broadinstitute/regional_missense_constraint/blob/main/rmc/pipeline/regional_constraint.py#L455); a breakpoint was likely at an exon boundary, so we unintentionally made the next section start in the intron

rmc/utils/constraint.py
lw453: Ah it is an array when you collect in the aggregation below this (which, you're right, should be fixed in the schema in the function documentation) but here this is a row-wise annotation summarizing the row which corresponds to a single region right?

ch-kr: oh good point, will update!

rmc/utils/generic.py
```diff
@@ -289,50 +289,68 @@ def get_ref_aa(
...
-    )
+        aa_map.update(extra_aa_map)
+    aa_map = hl.literal(aa_map)
+    ht = ht.annotate(ref_aa=aa_map.get(ht.ref_aa_1_letter, ht.ref_aa_1_letter))
     if aa_to_remove:
-        ht = ht.filter(ht.ref_aa != aa_to_remove)
+        ht = ht.filter(hl.literal(aa_to_remove).contains(ht.ref_aa))
+
+    # Select fields and checkpoint
+    ht = ht.select("ref_aa", "aa_start_num", "transcript")
```
lw453, line 306:
```suggestion
    ht = ht.select("ref_aa", "aa_start_num", "aa_end_num", "transcript")
```

rmc/utils/generic.py
```diff
@@ -289,50 +289,68 @@ def get_ref_aa(
...
+    ref_aa_check = ref_aa_check.difference(set(hl.eval(aa_map).keys()))
+    if len(ref_aa_check) != 0:
+        logger.warning(
+            "The following reference amino acids were not mapped to three letter"
+            " codes: %s",
+            ref_aa_check,
+        )
 
     # Double check that protein start always equals protein end
     protein_num_check = ht.aggregate(
```
lw453, line 324: Should we maybe make this check and the one on reference AAs into an expression that we write out + read if exists? Like in the LOFTEE aggs of misbad calculation. So we don't have to recompute every time we run.

rmc/utils/generic.py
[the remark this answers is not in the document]
lw453: Ok, if you're not worried about the cost!

rmc/utils/generic.py
```diff
@@ -289,50 +289,68 @@ def get_ref_aa(
...
-    ht = ht.checkpoint(
-        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/ref_amino_acids.ht",
-        _read_if_exists=not overwrite_temp,
-        overwrite=overwrite_temp,
-    )
+    ht = ht.annotate(ref_aa=hl.format("%s%s", ht.ref_aa, ht.aa_start_num))
 
     # Collect by key to collapse AA per locus
-    ht = ht.key_by("locus", "transcript").drop("alleles")
+    ht = ht.key_by("locus", "transcript").drop("alleles", "aa_start_num")
```
lw453, line 336:
```suggestion
    ht = ht.key_by("locus", "transcript").drop("alleles", "aa_start_num", "aa_end_num")
```

rmc/utils/generic.py
```diff
@@ -289,50 +289,68 @@ def get_ref_aa(
...
     # Add any extra mappings into amino acid mapping dict
     if extra_aa_map:
-        aa_map = aa_map.update(extra_aa_map)
-    ht = ht.annotate(ref_aa_1_letter=ht.amino_acids.split("/")[0])
-    ht = ht.annotate(
-        ref_aa=hl.literal(aa_map).get(ht.ref_aa_1_letter, ht.ref_aa_1_letter)
-    )
+        aa_map.update(extra_aa_map)
+    aa_map = hl.literal(aa_map)
+    ht = ht.annotate(ref_aa=aa_map.get(ht.ref_aa_1_letter, ht.ref_aa_1_letter))
```
lw453, line 301: `ht` actually has the field `amino_acids`, not `ref_aa_1_letter`

rmc/utils/generic.py
```diff
@@ -289,50 +289,68 @@ def get_ref_aa(
...
+    ht = ht.select("ref_aa", "aa_start_num", "transcript")
+    ht = ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/ref_amino_acids.ht",
+        _read_if_exists=not overwrite_temp,
+        overwrite=overwrite_temp,
+    )
+
+    # Check if any there are any ref amino acids in HT that aren't in aa_map
+    ref_aa_check = ht.aggregate(hl.agg.collect_as_set(ht.ref_aa))
+    ref_aa_check = ref_aa_check.difference(set(hl.eval(aa_map).keys()))
```
lw453, line 315:
```suggestion
    ref_aa_check = ref_aa_check.difference(set(hl.eval(aa_map).values()))
```

rmc/utils/generic.py
```diff
@@ -289,50 +289,68 @@ def get_ref_aa(
...
     ht = ht.collect_by_key(name="aa_info")
     ht = ht.checkpoint(
         f"{TEMP_PATH_WITH_FAST_DEL}/rmc/ref_aa_collected.ht",
         _read_if_exists=not overwrite_temp,
         overwrite=overwrite_temp,
     )
+
+    # Check to see if AA info is defined for all alleles associated with a locus/transcript
+    ht = ht.annotate(
+        all_aa_def=hl.all(hl.map(lambda x: hl.is_missing(x.ref_aa), ht.aa_info))
```
lw453, line 346:
```suggestion
        all_aa_def=hl.all(hl.map(lambda x: hl.is_defined(x.ref_aa), ht.aa_info))
```

rmc/utils/generic.py
```diff
@@ -289,50 +289,68 @@ def get_ref_aa(
...
-    ht = ht.key_by("locus", "transcript").drop("alleles")
+    ht = ht.key_by("locus", "transcript").drop("alleles", "aa_start_num")
     ht = ht.collect_by_key(name="aa_info")
     ht = ht.checkpoint(
         f"{TEMP_PATH_WITH_FAST_DEL}/rmc/ref_aa_collected.ht",
         _read_if_exists=not overwrite_temp,
         overwrite=overwrite_temp,
     )
+
+    # Check to see if AA info is defined for all alleles associated with a locus/transcript
```
lw453, line 344:
```suggestion
    # Check to see if AA info is defined for all alleles associated with a locus-transcript combination
```

rmc/utils/generic.py
```diff
@@ -289,50 +289,68 @@ def get_ref_aa(
...
-    ht = ht.annotate(ref_aa=hl.format("%s%s", ht.ref_aa, ht.protein_start))
-
-    # Select fields and checkpoint
-    ht = ht.select("ref_aa", "transcript")
-    ht = ht.checkpoint(
-        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/ref_amino_acids.ht",
-        _read_if_exists=not overwrite_temp,
-        overwrite=overwrite_temp,
-    )
+    ht = ht.annotate(ref_aa=hl.format("%s%s", ht.ref_aa, ht.aa_start_num))
```
lw453, line 333: Won't this produce `"null"` in place of `%s` if either `ref_aa` or `aa_start_num` is NA? So could be `"nullnull"` and would still register as a valid defined AA label in the check on line 346

rmc/utils/generic.py
```diff
@@ -289,50 +289,68 @@ def get_ref_aa(
...
+
+    # Check to see if AA info is defined for all alleles associated with a locus/transcript
+    ht = ht.annotate(
+        all_aa_def=hl.all(hl.map(lambda x: hl.is_missing(x.ref_aa), ht.aa_info))
+    )
+    missing_aa_check = ht.aggregate(hl.agg.count_where(~ht.all_aa_def))
+    if missing_aa_check != 0:
+        logger.warning(
+            "Found that %i amino acids were missing! (Some alleles for a"
+            " locus/transcript combination had missing AA information!)"
```
lw453, line 352:
```suggestion
            "%i locus-transcript combinations had missing AA info for at least 1 allele!"
```

rmc/utils/constraint.py
```diff
@@ -1533,35 +1554,45 @@ def check_and_fix_missing_aa(
 
     # Get CDS start/stops from CDS HT
     transcript_ht = transcript_cds.ht().key_by()
```
lw453, line 1556: Same as a previous comment — use `transcript_ref` instead

rmc/utils/constraint.py
```diff
@@ -1353,65 +1357,70 @@ def join_and_fix_aa(ht: hl.Table, fix_ht: hl.Table) -> hl.Table:
...
         If False, will read existing temp data rather than overwriting.
         If True, will overwrite temp data.
     :return: HT containing start and stop amino acids only for RMC regions that were
         previously missing amino acid information.
     """
+    # Filter to regions missing AA at transcript starts/ends
     miss_start_stop_ht = ht.filter(
         (hl.is_missing(ht.start_aa) & ht.transcript_start)
         | (hl.is_missing(ht.stop_aa) & ht.transcript_stop)
     )
```
lw453:
```suggestion
        (hl.is_missing(ht.start_aa) & ht.is_transcript_start)
        | (hl.is_missing(ht.stop_aa) & ht.is_transcript_stop)
    )
```

rmc/utils/constraint.py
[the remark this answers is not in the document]
lw453: Hm why is this the best choice? The closest AA isn't necessarily the right one

rmc/utils/constraint.py
lw453: So sounds like this functionality will become obsolete if we choose to not include introns at the ends of regions

rmc/utils/constraint.py
```diff
@@ -1353,65 +1357,70 @@ def join_and_fix_aa(ht: hl.Table, fix_ht: hl.Table) -> hl.Table:
...
-    filt_context_ht = get_aa_from_context(
-        overwrite_temp=False, keep_transcripts=miss_start_stop_transcripts
+    context_ht = context_ht.filter(
+        hl.literal(miss_start_stop_transcripts).contains(context_ht.transcript)
     )
-    filt_context_ht = filt_context_ht.filter(
-        hl.literal(miss_start_stop_transcripts).contains(filt_context_ht.transcript)
+    context_ht = context_ht.group_by("transcript").aggregate(
+        max_aa_num=hl.agg.max(context_ht.protein_start)
```
lw453, line 1415:
```suggestion
        max_aa_num=hl.agg.max(context_ht.aa_start_num)
```

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+        **filt_context_ht[miss_start_stop_ht.transcript]
+    )
+    miss_start_stop_ht = miss_start_stop_ht.annotate(
+        start_aa=hl.if_else(
+            hl.is_missing(miss_start_stop_ht.start_aa),
+            hl.if_else(
+                # Set all missing + strand transcript starts to Met1
+                miss_start_stop_ht.strand == "+",
+                "Met1",
+                hl.format("Ter%s", miss_start_stop_ht.max_aa_num),
```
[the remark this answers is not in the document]
lw453: For other transcripts that don't have this start/stop issue, isn't it an actual AA that ends it?

rmc/utils/constraint.py
lw453: No I mean, this function as it is only adjusts start/stop AAs if they're at the start/end of transcripts but you use it in line 1584 to adjust region start/stop AAs too.

lw453 (changes requested):

rmc/pipeline/regional_constraint.py
```diff
@@ -303,8 +304,19 @@ def main(args):
                     logger.info(
                         "Adding section annotation before searching for first break..."
                     )
-                    # Add transcript start and stop positions from browser HT
-                    transcript_ht = gene_model.ht().select("start", "stop")
+                    # Add transcript start and stop positions from CDS HT
+                    # NOTE: For RMC freezes 1-7, the transcript start and ends were from the
+                    # browser HT (`gene_model`) whose start and stop generally is in the UTRs
+                    transcript_ht = transcript_cds.ht()
```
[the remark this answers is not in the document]
ch-kr: oh wow, I completely missed this. thanks!

rmc/utils/generic.py
ch-kr: oh no, this must have gotten dropped when I applied a suggestion in vscode

rmc/utils/generic.py
ch-kr: good idea, done

rmc/utils/constraint.py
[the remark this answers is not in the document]
ch-kr: yup

rmc/utils/constraint.py
ch-kr: yep!

rmc/utils/constraint.py
ch-kr: oh shoot, yes

rmc/utils/constraint.py
```diff
@@ -1310,45 +1318,350 @@ def create_context_with_oe(
...
+    missing_start_or_stop = missing_ht.count()
+    if missing_start_or_stop == 0:
+        return ht
+
+    logger.info(
+        "%i sites are missing either their stop or start AA!", missing_start_or_stop
+    )
+
+    # Get strand from browser HT
+    # TODO: recreate browser HT to include strand annotation
```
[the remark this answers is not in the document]
ch-kr: actually I will remake this now

rmc/utils/generic.py
```diff
@@ -237,6 +237,166 @@ def process_context_ht(
...
+            missing_aa_check, missing_aa_check_he_path, overwrite=True
+        )
+
+    missing_aa_check = hl.eval(
+        hl.experimental.read_expression(missing_aa_check_he_path)
+    )
+    if missing_aa_check != 0:
+        logger.warning(
+            "%i locus-transcript combinations had missing AA info for at least 1"
+            " allele!"
```
Ruchit10, line 395:
```suggestion
            " allele!", missing_aa_check
```

rmc/utils/constraint.py
```diff
@@ -380,9 +389,16 @@ def create_constraint_prep_ht(
...
 
     logger.info("Adding section annotation...")
-    # Add transcript start and stop positions from browser HT
-    transcript_ht = gene_model.ht().select("start", "stop")
-    ht = ht.annotate(**transcript_ht[ht.transcript])
+    # Add transcript start and stop CDS positions
+    # NOTE: RMC freezes 1-7 used `gene_model` resource to get transcript start and stops
+    # rather than CDS
+    transcript_ht = (
+        transcript_ref.ht().ht().select("gnomad_cds_start", "gnomad_cds_end")
```
Ruchit10, line 396: Why are there two `.ht()`?

Ruchit10 (changes requested):
Some minor comments, otherwise looks good

rmc/utils/generic.py
```diff
@@ -237,6 +237,166 @@ def process_context_ht(
...
+    :param extra_aa_map: Dictionary mapping any amino acid one letter codes to three letter codes.
+        Designed to capture any amino acids not present in `ACID_NAMES_PATH`.
+        Default is {"*": "Ter", "U": "Sec"}.
+    :param aa_to_remove: Any amino acid one letter codes to remove. Default is {"X"}.
+    :return: HT with one reference amino acid annotated per locus/transcript.
+    """
+    logger.info("Mapping amino acid one letter codes to three letter codes...")
+    aa_map = get_aa_map()
+    # Add any extra mappings into amino acid mapping dict
+    if extra_aa_map:
```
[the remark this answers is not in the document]
ch-kr: that's true, but this seemed more flexible than either having an additional if check to avoid running this update if `extra_aa_map` is None / raising an error if `extra_aa_map` is None. what do you think?

rmc/utils/generic.py
```diff
@@ -237,6 +237,166 @@ def process_context_ht(
...
+        aa_map.update(extra_aa_map)
+    aa_map = hl.literal(aa_map)
+    ht = ht.annotate(ref_aa_1_letter=ht.amino_acids.split("/")[0])
+    ht = ht.annotate(ref_aa=aa_map.get(ht.ref_aa_1_letter, ht.ref_aa_1_letter))
+    if aa_to_remove:
```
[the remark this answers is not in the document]
ch-kr: same comment here

rmc/utils/constraint.py
ch-kr: probably an error introduced when fixing merge conflicts -- thanks for flagging!

rmc/utils/generic.py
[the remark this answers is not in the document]
Ruchit10: makes sense. No harm in keeping this logic. Just wondering if this logic was deliberate which it is, so all good :)

rmc/utils/generic.py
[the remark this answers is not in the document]
Ruchit10: All good!

Ruchit10 (approved):
Lily is going to re-review this but looks good to me!

rmc/utils/constraint.py
```diff
@@ -1454,63 +1548,110 @@ def fix_region_start_stop_aas(
...
+    def _create_exon_position_ht(
+        cds_ht: hl.Table, fix_missing_start: bool, overwrite_temp: bool
+    ) -> hl.Table:
+        """
+        Create Table keyed by either exon start or stop position.
+
+        :param cds_ht: CDS HT resource filtered to transcripts missing region start or stop
+            AA annotations only.
+        :param fix_missing_start: Whether the output HT is designed to fix region starts
+            that are missing AA annotations.
```
lw453, line 1580:
```suggestion
            that are missing AA annotations.            
            If True, will return the Table keyed by the exon stop and annotated with the next exon start.
            If False, will return the Table keyed by the exon start and annotated with the previous exon stop.
```

rmc/utils/constraint.py
```diff
@@ -1454,63 +1548,110 @@ def fix_region_start_stop_aas(
...
+            checkpoint_path = f"{TEMP_PATH_WITH_FAST_DEL}/cds_start_fix.ht"
+            cds_ht = cds_ht.order_by(
+                hl.asc(cds_ht.transcript), hl.desc(cds_ht.exon_start)
+            )
+            cds_ht = cds_ht.annotate(
+                next_exon_start=hl.scan._prev_nonnull(cds_ht.exon_start)
+            )
+            cds_ht = cds_ht.key_by(cds_ht.exon_stop, cds_ht.transcript)
+        else:
+            checkpoint_path = f"{TEMP_PATH_WITH_FAST_DEL}/cds_stop_fix.ht"
```
lw453, line 1599: I think a reorder here could possibly be needed just to ensure correctness since the original key is interval then transcript
```suggestion
            checkpoint_path = f"{TEMP_PATH_WITH_FAST_DEL}/cds_stop_fix.ht"
            cds_ht = cds_ht.order_by("transcript", "exon_stop")
```

rmc/utils/constraint.py
```diff
@@ -1454,63 +1548,110 @@ def fix_region_start_stop_aas(
...
+            AA annotations only.
+        :param fix_missing_start: Whether the output HT is designed to fix region starts
+            that are missing AA annotations.
+        :param overwrite_temp: Whether to overwrite temporary data.
+            If False, will read existing temp data rather than overwriting.
+            If True, will overwrite temp data.
+        :return: CDS HT keyed by either exon start or stop position and transcript.
+        """
+        if fix_missing_start:
+            # Region starts missing AAs need to get adjusted to use the AA from the previous
```
lw453, line 1587:
```suggestion
            # Region starts missing AAs need to get adjusted to use the AA from the next
```

rmc/utils/constraint.py
```diff
@@ -1454,63 +1548,110 @@ def fix_region_start_stop_aas(
...
-            hl.is_defined(missing_ht.start_aa),
-            hl.locus(missing_ht.start_coordinate.contig, missing_ht.closest_start_mis),
-            missing_ht.start_coordinate,
+    stop_fix_ht = _create_exon_position_ht(
+        cds_ht=cds_ht, fix_missing_start=False, overwrite_temp=overwrite_temp
+    )
+    missing_ht = missing_ht.annotate(
+        next_exon_start=hl.or_missing(
+            hl.is_missing(missing_ht.start_aa),
+            stop_fix_ht[
```
lw453, line 1622: I think these are switched
```suggestion
            start_fix_ht[
```

rmc/utils/constraint.py
```diff
@@ -1454,63 +1548,110 @@ def fix_region_start_stop_aas(
...
+                missing_ht.transcript,
+            ].next_exon_start,
         ),
-        stop_coordinate=hl.if_else(
-            hl.is_defined(missing_ht.stop_aa),
-            hl.locus(missing_ht.start_coordinate.contig, missing_ht.closest_stop_mis),
-            missing_ht.stop_coordinate,
+        prev_exon_stop=hl.or_missing(
+            hl.is_missing(missing_ht.stop_aa),
+            start_fix_ht[
```
lw453, line 1632:
```suggestion
            stop_fix_ht[
```

rmc/utils/constraint.py
```diff
@@ -1454,63 +1548,110 @@ def fix_region_start_stop_aas(
...
+            # Reorder HT so that _prev_nonnull returns the next exon start
+            checkpoint_path = f"{TEMP_PATH_WITH_FAST_DEL}/cds_start_fix.ht"
+            cds_ht = cds_ht.order_by(
+                hl.asc(cds_ht.transcript), hl.desc(cds_ht.exon_start)
+            )
+            cds_ht = cds_ht.annotate(
+                next_exon_start=hl.scan._prev_nonnull(cds_ht.exon_start)
+            )
+            cds_ht = cds_ht.key_by(cds_ht.exon_stop, cds_ht.transcript)
+        else:
```
lw453, line 1598:
```suggestion
        else:
            # Region stops missing AAs need to get adjusted to use the AA from the previous
            # exon stop
```

lw453 (changes requested):

rmc/utils/constraint.py
```diff
@@ -1305,44 +1319,534 @@ def create_context_with_oe(
...
+    # Annotate amino acids
+    ht = annot_rmc_with_aa(ht, overwrite_temp)
+
+    # Remove missense O/E cap of 1
+    # (Missense O/E capped for RMC search, but
+    # it shouldn't be capped when displayed in the browser)
+    ht = ht.annotate(section_oe=ht.obs / ht.exp)
+
+    # Convert chi square to p-value
+    ht = ht.annotate(section_p_value=hl.dchisq(ht.section_chisq, 1))
```
lw453, line 1828: `dchisq` is calculating the probability density at that specific chisq value in the distribution, we want `pchisqtail` for the cumulative probability under the distribution at this chi square value or greater.
```suggestion
    ht = ht.annotate(section_p_value=hl.pchisqtail(ht.section_chisq, 1))
```

lw453 (changes requested):

rmc/utils/generic.py
```diff
@@ -338,19 +339,35 @@ def get_ref_aa(
...
+        # even if file already exists
+        protein_num_check = ht.aggregate(
+            hl.agg.count_where(ht.aa_start_num != ht.aa_end_num)
+        )
+        hl.experimental.write_expression(
+            protein_num_check, protein_num_check_he_path, overwrite=True
+        )
+
+    protein_num_check = hl.eval(
+        hl.experimental.read_expression(protein_num_check_he_path)
```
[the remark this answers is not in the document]
ch-kr: hmm I don't see why this is better? this suggestion would have to be:
```
    if not overwrite_temp and file_exists(protein_num_check_he_path):
        protein_num_check = hl.eval(
            hl.experimental.read_expression(protein_num_check_he_path)
        )
    else:
        protein_num_check = ht.aggregate(
            hl.agg.count_where(ht.aa_start_num != ht.aa_end_num)
        )
        hl.experimental.write_expression(
            protein_num_check, protein_num_check_he_path, overwrite=overwrite_temp
        )
        protein_num_check = hl.eval(
            hl.experimental.read_expression(protein_num_check_he_path)
        )
```
to work, right?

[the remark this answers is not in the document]
ch-kr: ah, is the suggestion is to avoid the unnecessary aggregation if this does already exist?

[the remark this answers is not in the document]
ch-kr: will change to
```
    # Double check that protein start always equals protein end
    protein_num_check_he_path = f"{TEMP_PATH_WITH_FAST_DEL}/protein_num_count.he"
    if not file_exists(protein_num_check_he_path):
        protein_num_check = ht.aggregate(
            hl.agg.count_where(ht.aa_start_num != ht.aa_end_num)
        )
        hl.experimental.write_expression(
            protein_num_check, protein_num_check_he_path, overwrite=True
        )

    # Assume file already exists otherwise
    protein_num_check = hl.eval(
        hl.experimental.read_expression(protein_num_check_he_path)
    )
```

rmc/utils/generic.py
```diff
@@ -359,14 +376,24 @@ def get_ref_aa(
...
+            all_aa_def=hl.all(hl.map(lambda x: hl.is_missing(x.ref_aa), ht.aa_info))
+        )
+        missing_aa_check = ht.aggregate(hl.agg.count_where(ht.all_aa_def))
+        missing_aa_check = hl.experimental.write_expression(
+            missing_aa_check, missing_aa_check_he_path, overwrite=True
+        )
+
+    missing_aa_check = hl.eval(
+        hl.experimental.read_expression(missing_aa_check_he_path)
     )
```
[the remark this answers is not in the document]
ch-kr: wait, for the `~`:
```
        ht = ht.annotate(
            all_aa_def=hl.all(hl.map(lambda x: hl.is_missing(x.ref_aa), ht.aa_info))
        )
```
is first annotating a list of booleans based on whether the AA was missing (so `[False, False, False]` if all AA were defined) and then flattening it using `hl.all` to be a single boolean (so a single `False` in this case). the aggregate then counts any instances of True, which would indicate an AA was actually missing in the list, so not adding the `~` is the correct behavior, right? maybe the fix is actually renaming `all_aa_def` -> `any_aa_missing`?

rmc/utils/generic.py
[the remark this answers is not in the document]
lw453: Oh I thought we would write it regardless if `overwrite_temp` is True?

rmc/utils/generic.py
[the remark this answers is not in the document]
lw453: Oh yes, agree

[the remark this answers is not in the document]
lw453: Wait. Don't we want `all_aa_def=hl.all(hl.map(lambda x: hl.is_defined(x.ref_aa), ht.aa_info))` and then `missing_aa_check = ht.aggregate(hl.agg.count_where(~ht.all_aa_def))`? Because `hl.all` is only True if all the elements are True. So any instances of `hl.all=True` will only arise if all AAs are missing in the list

rmc/utils/constraint.py
ch-kr: BAHA I always miss copy paste errors, thank you for catching

rmc/utils/generic.py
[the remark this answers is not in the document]
lw453: Rather, I meant that if the file already exists but we want to overwrite it so `overwrite_temp` is True, your suggestion won't overwrite

rmc/utils/generic.py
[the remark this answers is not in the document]
lw453: Lol yeah that works. So no `~`

rmc/utils/generic.py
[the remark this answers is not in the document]
ch-kr: oh merp, you're right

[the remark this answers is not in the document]
ch-kr: fixed!

rmc/utils/generic.py
```diff
@@ -237,6 +237,181 @@ def process_context_ht(
...
+
+    # Assume file already exists otherwise
+    protein_num_check = hl.eval(
+        hl.experimental.read_expression(protein_num_check_he_path)
+    )
+    if protein_num_check != 0:
+        raise DataException(
+            f"{protein_num_check} sites had different amino acid numbers at start and"
+            " end -- please double check!"
+        )
```
lw453, line 371:
```suggestion
        )
        
```

rmc/utils/generic.py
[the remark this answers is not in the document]
lw453: The new structure is fine with me and looks cleaner, but there are places where we have the old structure in the code e.g. for the LoFTEE agg

rmc/utils/generic.py
```diff
@@ -377,11 +388,14 @@ def get_ref_aa(
 
     # Check to see if AA info is defined for all alleles associated with a locus/transcript
     missing_aa_check_he_path = f"{TEMP_PATH_WITH_FAST_DEL}/missing_aa_check.he"
-    if overwrite_temp:
+    overwrite_he = (
+        file_exists(missing_aa_check_he_path) if not overwrite_temp else overwrite_temp
```
lw453, line 392:
```suggestion
        not file_exists(missing_aa_check_he_path) if not overwrite_temp else overwrite_temp
```

rmc/utils/constraint.py
```diff
@@ -1354,40 +1354,23 @@ def annot_rmc_with_aa(ht: hl.Table, overwrite_temp: bool):
...
-            stop_aa=hl.if_else(
-                hl.is_missing(ht.stop_aa) & ht.is_transcript_stop,
-                fix_ht[ht.interval, ht.transcript].stop_aa,
-                ht.stop_aa,
-            ),
-        )
     return ht.annotate(
         start_aa=hl.if_else(
-            hl.is_missing(ht.start_aa),
+            hl.is_missing(ht.start_aa) & ht.is_transcript_start,
```
lw453, line 1368:
```suggestion
            hl.is_missing(ht.start_aa),
```

rmc/utils/constraint.py
```diff
@@ -1354,40 +1354,23 @@ def annot_rmc_with_aa(ht: hl.Table, overwrite_temp: bool):
...
             fix_ht[ht.interval, ht.transcript].start_aa,
             ht.start_aa,
         ),
         stop_aa=hl.if_else(
-            hl.is_missing(ht.stop_aa),
+            hl.is_missing(ht.stop_aa) & ht.is_transcript_stop,
```
lw453, line 1373:
```suggestion
            hl.is_missing(ht.stop_aa),
```

rmc/utils/constraint.py
```diff
@@ -1527,11 +1490,13 @@ def fix_region_start_stop_aas(
...
         - This function assumes that all start coordinates missing AA annotations are one position larger than
-            the previous exon stop position. (This was the case in RMC freeze 5.)
+            the previous exon stop position. (This was the case in RMC freeze 7.)
+            This means that the real start coordinate should be at the next exon start.
         - This function assumes that all stop coordinates missing AA annotations are one position smaller than
-            the following exon start position. (This was the case in RMC freeze 5).
-        - See this ticket for more information about RMC freeze 5:
-            https://github.com/broadinstitute/rmc_production/issues/149
+            the following exon start position. (This was the case in RMC freeze 7.)
+            This means that the real start coordinate should be at the previous exon stop.
```
lw453, line 1497: I messed up lol
```suggestion
            This means that the real stop coordinate should be at the previous exon stop.
```

rmc/utils/constraint.py
```diff
@@ -1447,44 +1430,36 @@ def fix_transcript_start_stop_aas(
         overwrite=overwrite_temp,
     )
 
-    # Get all positions with defined amino acids in context HT
+    # Get largest defined amino acids present in context HT for each
+    # transcript that is missing an amino acid at its CDS start or end
```
lw453, line 1434: I think we can just move this to line 1445 (+ replace the current comment there)

rmc/utils/constraint.py
```diff
@@ -1447,44 +1430,36 @@ def fix_transcript_start_stop_aas(
...
-        # len(pos_ht[miss_start_stop_ht.transcript].positions)
-        # Therefore taking the last index of the list and skipping the binary search
-        closest_stop_pos=pos_ht[miss_start_stop_ht.transcript].positions[-1],
-    )
-    # Checkpoint to make sure the binary search/join only runs once,
-    # especially since there is another join immediately afterwards
+        largest_aa=max_aa_ht[miss_start_stop_ht.transcript].largest_aa,
+    )
     miss_start_stop_ht = miss_start_stop_ht.checkpoint(
         f"{TEMP_PATH_WITH_FAST_DEL}/ts_missing_start_stop_pos.ht",
```
lw453, line 1464:
```suggestion
        f"{TEMP_PATH_WITH_FAST_DEL}/ts_missing_start_stop_largest.ht",
```

rmc/utils/constraint.py
```diff
@@ -1596,7 +1565,12 @@ def _create_exon_position_ht(
             )
             cds_ht = cds_ht.key_by(cds_ht.exon_stop, cds_ht.transcript)
         else:
+            # Reorder HT to ensure that the correct exon stop number is returned
+            # (the previous key is interval and transcript)
+            # Region stops missing AAsneed to get adjusted to use the AA from the
```
lw453, line 1570:
```suggestion
            # Region stops missing AAs need to get adjusted to use the AA from the
```

lw453 (changes requested):
Sorry I accidentally submitted the first half of my review half-way through

rmc/utils/constraint.py
```diff
@@ -1319,15 +1319,15 @@ def create_context_with_oe(
...
     """
-    Annotate RMC results HT with amino acid information.
+    Annotate RMC regions HT with amino acids at region starts and stops.
 
     :param ht: Input RMC results HT.
     :param overwrite_temp: Whether to overwrite temporary data.
         If False, will read existing temp data rather than overwriting.
         If True, will overwrite temp data.
-    :return: RMC results HT annotated with amino acid information.
+    :return: RMC results HT annotated with amino acid information for region starts and stops.
```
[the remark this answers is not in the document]
ch-kr: done, let me know if I missed any!

rmc/utils/generic.py
```diff
@@ -237,6 +237,181 @@ def process_context_ht(
...
+            hl.is_defined(ht.ref_aa) & hl.is_defined(ht.aa_start_num),
+            hl.format("%s%s", ht.ref_aa, ht.aa_start_num),
+        )
+    )
+
+    # Collect by key to collapse AA per locus
+    ht = ht.key_by("locus", "transcript").drop("alleles", "aa_start_num", "aa_end_num")
+    ht = ht.collect_by_key(name="aa_info")
+    ht = ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc/ref_aa_collected.ht",
```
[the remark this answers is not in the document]
ch-kr: will add locally -- looks like pylint is complaining about formatting in this script

rmc/utils/constraint.py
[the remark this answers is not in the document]
lw453: Line 1395 and 1641?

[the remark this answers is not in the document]
ch-kr: thanks!

lw453 (approved):

PR #301, opened 2024-01-17
Deduplicate MPC prior to liftover

ch-kr:
- Created new function `dedup_annot` to deduplicate OE annotation when creating `context_with_oe` resource and MPC when lifting over MPC release
- Minor clean up of `generic.py`; removed unused code and added missing section header
- Minor improvements to `liftover_mpc`: rename transcript to reflect that this was the GRCh37 transcript ID and remove unnecessary field `locus_fail_liftover` from liftover MPC release

### DIFF
--- a/rmc/utils/constraint.py
+++ b/rmc/utils/constraint.py
@@ -1257,7 +1257,7 @@ def dedup_annot(
     :param keep_transcript: Whether to retain transcript ID.
     :param key_fields: Key fields to use for deduplication.
     """
-    ht = ht.select(annot).key_by(*(key_fields))
+    ht = ht.key_by(*(key_fields)).select(annot)
     ht = ht.collect_by_key()
 
     if get_min:
--- a/rmc/utils/constraint.py
+++ b/rmc/utils/constraint.py
@@ -1,7 +1,7 @@
 import logging
 import re
 import subprocess
-from typing import List, Set, Union
+from typing import List, Set, Tuple, Union
 
 import hail as hl
 import scipy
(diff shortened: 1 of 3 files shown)


rmc/utils/constraint.py
```diff
@@ -1239,6 +1239,38 @@ def get_oe_annotation(ht: hl.Table, freeze: int) -> hl.Table:
...
+    Function will group input Table by `key_fields`, select either the min or max value of
+    `annot`, and optionally retain transcript ID.
+
+    :param ht: Input Table.
+    :param annot: Name of annotation to be deduplicated.
+    :param get_min: Whether to get minimum value of annotation.
+    :param keep_transcript: Whether to retain transcript ID.
+    :param key_fields: Key fields to use for deduplication.
+    """
+    ht = ht.select(annot).key_by(*(key_fields))
```
Ruchit10, line 1260: Doesn't this already assume HT is keyed by `key_fields`? If not keyed, then `select` would drop them. Hope I'm getting it right

ch-kr: yes, good point!

rmc/utils/mpc.py
```diff
@@ -1062,5 +1067,18 @@ def liftover_mpc(
             " to GRCh38."
         )
     ht = hl.read_table(mpc_release.versions[freeze].path)
+    # Deduplicate MPC annotation by keeping largest MPC value per variant (locus/alleles) combination
+    ht = dedup_annot(ht, annot="mpc", get_min=False)
+    # Rename transcript information
+    ht = ht.transmute(transcript_grch37=ht.transcript)
```
Ruchit10, line 1073: Curious if `ht.rename` would be faster/slower or same?

ch-kr: I'm not sure, I've never used rename. transmute seems a little more straightforward

Ruchit10: either should work, I was just curious :)

Ruchit10 (approved):
Looks Good to me

PR #304, opened 2024-02-13
Add validity checks for RMC release

ch-kr:
Add new functions to validity check RMC release (to prevent https://github.com/broadinstitute/rmc_production/issues/200 from happening again)

### DIFF
--- a/rmc/utils/constraint.py
+++ b/rmc/utils/constraint.py
@@ -1596,16 +1596,16 @@ def fix_region_start_stop_aas(
             cds_ht = cds_ht.annotate(
                 next_exon_start=hl.scan._prev_nonnull(cds_ht.exon_start)
             )
-            cds_ht = cds_ht.key_by(cds_ht.exon_stop, cds_ht.transcript)
+            cds_ht = cds_ht.key_by(cds_ht.exon_end, cds_ht.transcript)
         else:
             # Reorder HT to ensure that the correct exon stop number is returned
             # (the previous key is interval and transcript)
             # Region stops missing AAs need to get adjusted to use the AA from the
             # previous exon stop
             checkpoint_path = f"{TEMP_PATH_WITH_FAST_DEL}/cds_stop_fix.ht"
-            cds_ht = cds_ht.order_by("transcript", "exon_stop")
+            cds_ht = cds_ht.order_by("transcript", "exon_end")
             cds_ht = cds_ht.annotate(
-                prev_exon_stop=hl.scan._prev_nonnull(cds_ht.exon_stop)
+                prev_exon_end=hl.scan._prev_nonnull(cds_ht.exon_end)
             )
             cds_ht = cds_ht.key_by(cds_ht.exon_start, cds_ht.transcript)
 
--- a/rmc/utils/constraint.py
+++ b/rmc/utils/constraint.py
@@ -2002,11 +2002,11 @@ def check_for_overlapping_intervals(interval_ht: hl.Table, coord_ht: hl.Table) -
         impacted_transcripts = coord_ht.aggregate(
             hl.agg.collect_as_set(coord_ht.transcript)
         )
-        logger.info(
+        logger.warning(
             "Number of transcripts with overlapping regions: %i",
             len(impacted_transcripts),
         )
-        logger.info("Transcripts with overlapping regions: %s", impacted_transcripts)
+        logger.warning("Transcripts with overlapping regions: %s", impacted_transcripts)
 
 
 def import_tsv_and_agg_transcripts(tsv_path: str) -> Set[str]:
--- a/rmc/utils/constraint.py
+++ b/rmc/utils/constraint.py
@@ -1563,7 +1563,7 @@ def fix_region_start_stop_aas(
     cds_ht = cds_ht.filter(missing_transcripts.contains(cds_ht.transcript))
     cds_ht = cds_ht.annotate(
         exon_start=cds_ht.interval.start,
-        exon_end=cds_ht.interval.end,
+        exon_stop=cds_ht.interval.end,
     )
 
     # Create two versions of CDS HT to fix missing region start and stop AAs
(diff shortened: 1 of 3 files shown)


rmc/utils/constraint.py
```diff
@@ -1708,10 +1709,13 @@ def _missing_aa_check(ht: hl.Table) -> Union[hl.Table, None]:
         return ht
 
     # Get CDS start/stops from CDS HT
-    transcript_ht = transcript_ref.ht().key_by()
+    transcript_ht = transcript_ref.ht().key_by("transcript")
```
lw453, line 1712: The table is already keyed by transcript

rmc/utils/constraint.py
```diff
@@ -1969,6 +1973,120 @@ def _annotate_gene_ids(ht: hl.Table) -> hl.Table:
...
+
+
+def validate_rmc_release_downloads(freeze: int) -> None:
+    """
+    Run validity checks on RMC downloadable files.
+
+    Function checks that:
+        - There are no overlapping intervals within transcripts
+        - Transcript set in RMC TSV matches transcript set in RMC HT
+        - Transcript set in no-RMC TSV does not match transcript set in RMC HT
```
lw453, line 2033:
```suggestion
        - Transcript set in no-RMC TSV has no matches with transcript set in RMC HT
```

rmc/utils/constraint.py
```diff
@@ -1969,6 +1973,120 @@ def _annotate_gene_ids(ht: hl.Table) -> hl.Table:
...
+    )
+
+    # Key one HT by interval, one by start coordinate, and one by end coordinate
+    ht = ht.key_by("interval")
+    start_ht = ht.key_by(locus=ht.start_coordinate)
+    end_ht = ht.key_by(locus=ht.stop_coordinate)
+
+    logger.info("Checking for overlapping intervals...")
+    logger.info("Checking if any start coordinates overlap more than 1 interval...")
+    check_for_overlapping_intervals(ht, start_ht)
```
lw453, line 2063:
```suggestion
    check_for_overlapping_intervals(ht, start_ht)
    logger.info("Checking if any end coordinates overlap more than 1 interval...")
```

rmc/utils/constraint.py
```diff
@@ -1969,6 +1973,120 @@ def _annotate_gene_ids(ht: hl.Table) -> hl.Table:
     )
 
 
+def check_for_overlapping_intervals(interval_ht: hl.Table, coord_ht: hl.Table) -> None:
+    """
+    Check for overlapping intervals within the same transcript.
```
lw453, line 1978:
```suggestion
    Check for overlapping regions/intervals within the same transcript.
```

rmc/utils/constraint.py
```diff
@@ -1969,6 +1973,120 @@ def _annotate_gene_ids(ht: hl.Table) -> hl.Table:
...
+    :param coord_ht: RMC HT keyed by either RMC interval start or end coordinate.
+    :return: None
+    """
+    coord_ht = coord_ht.annotate(
+        interval_matches=interval_ht.index(coord_ht.locus, all_matches=True),
+    )
+    coord_ht = coord_ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/rmc_coord_check.ht", overwrite=True
+    )
+    coord_ht = coord_ht.filter(hl.len(coord_ht.interval_matches) > 1)
```
lw453, line 1990: Line probably redundant since you've got line 1998

rmc/utils/constraint.py
```diff
@@ -1969,6 +1973,120 @@ def _annotate_gene_ids(ht: hl.Table) -> hl.Table:
...
+    # Keep only rows where start coordinate was found in two intervals **in the same transcript**
+    coord_ht = coord_ht.annotate(
+        interval_matches=coord_ht.interval_matches.filter(
+            lambda x: x.transcript == coord_ht.transcript
+        )
+    )
+    coord_ht = coord_ht.filter(hl.len(coord_ht.interval_matches) > 1)
+    n_overlap = coord_ht.count()
+    if n_overlap > 0:
+        coord_ht.show()
```
lw453, line 2001: Remove?

rmc/utils/constraint.py
```diff
@@ -1969,6 +1973,120 @@ def _annotate_gene_ids(ht: hl.Table) -> hl.Table:
...
+        impacted_transcripts = coord_ht.aggregate(
+            hl.agg.collect_as_set(coord_ht.transcript)
+        )
+        logger.info(
```
lw453, line 2005: Can make this and line below warnings?

rmc/utils/constraint.py
```diff
@@ -1483,14 +1483,15 @@ def fix_transcript_start_stop_aas(
...
+    max_aa_ht = max_aa_ht.key_by("transcript")
     max_aa_ht = max_aa_ht.annotate(
         largest_aa=max_aa_grp_ht[max_aa_ht.transcript].largest_aa,
     )
     max_aa_ht = max_aa_ht.filter(max_aa_ht.aa_num == max_aa_ht.largest_aa)
 
     # Annotate largest amino acid back onto original HT
     miss_start_stop_ht = miss_start_stop_ht.annotate(
-        largest_aa=max_aa_ht[miss_start_stop_ht.transcript].largest_aa,
+        largest_aa=max_aa_ht[miss_start_stop_ht.transcript].ref_aa,
```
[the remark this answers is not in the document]
ch-kr: that's a good question -- do you know if there are transcripts with missing strand types in the new build?

this question also reminds me that these checks built to handle only non-outlier transcripts and likely won't work when we expand to include all transcripts

rmc/utils/constraint.py
ch-kr: you're right -- this rekey may have been intended for `ht`

ch-kr: though it looks like that isn't the case below based on line 1719

rmc/utils/constraint.py
ch-kr: this may have been for efficiency -- I'll move it above the checkpoint

rmc/utils/constraint.py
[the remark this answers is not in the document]
lw453: I've got no idea about missing strands but I figure if a transcript has missing strand, we'll just have the AA stay missing - not worth it to try to infer from things like info on other AA in the middle. If there is an AA that stays missing, would you foresee any problems downstream?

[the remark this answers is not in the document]
ch-kr: oh, I see. keeping the AAs mising makes sense to me, and I don't think there would be issues with having missing values

rmc/utils/constraint.py
lw453: Oh ok - I thought it may have been a copy paste error from a notebook but if this is something you want to/can output for a check, sounds good!

lw453 (approved):

PR #312, opened 2025-04-02
Adapt RMC to use per variant expected counts from constraint table

Ruchit10:
Minor changes:

1. Comment out code not used which created per variant expected using mutation rates and context HT
2. Read output HT of `--apply-models` step of the constraint pipeline as input table for `prep-constraint` step of RMC pipeline

### DIFF
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -4,7 +4,6 @@ import logging
 import hail as hl
 from gnomad.utils.file_utils import file_exists
 
-
 from rmc.resources.basics import (
     CONSTRAINT_PREFIX,
     LOGGING_PATH,
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -26,7 +26,7 @@ from rmc.resources.rmc import (
     simul_search_round_bucket_path,
     single_search_round_ht_path,
 )
-from rmc.slack_creds import slack_token
+# from rmc.slack_creds import slack_token
 from rmc.utils.constraint import (
     annotate_max_chisq_per_section,
     check_break_search_round_nums,
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -26,6 +26,7 @@ from rmc.resources.rmc import (
     simul_search_round_bucket_path,
     single_search_round_ht_path,
 )
+
 # from rmc.slack_creds import slack_token
 from rmc.utils.constraint import (
     annotate_max_chisq_per_section,
@@ -65,20 +66,20 @@ def main(args):
             create_transcript_ref(build="GRCh38", overwrite=args.overwrite)
 
         # NOTE: This code block is no longer needed since the expected values are now calculated per-variant upstream
-        # if args.command == "prep-filtered-context":
-        #     hl.init(
-        #         log="/RMC_pre_process.log",
-        #         tmp_dir=TEMP_PATH_WITH_FAST_DEL,
-        #         quiet=args.quiet,
-        #     )
-        #     hl.default_reference("GRCh38")
-        #     logger.info("Creating filtered context HT...")
-        #     n_partitions = 10000
-        #     create_filtered_context_ht(
-        #         canonical_only=args.filter_to_canonical,
-        #         n_partitions=args.n_partitions if args.n_partitions else n_partitions,
-        #         overwrite=args.overwrite_temp,
-        #     )
+        if args.command == "prep-filtered-context":
+            hl.init(
+                log="/RMC_pre_process.log",
+                tmp_dir=TEMP_PATH_WITH_FAST_DEL,
+                quiet=args.quiet,
+            )
+            hl.default_reference("GRCh38")
+            logger.info("Creating filtered context HT...")
+            n_partitions = 10000
+            create_filtered_context_ht(
+                canonical_only=args.filter_to_canonical,
+                n_partitions=args.n_partitions if args.n_partitions else n_partitions,
+                overwrite=args.overwrite_temp,
+            )
 
         if args.command == "prep-constraint":
             hl.init(
@@ -643,8 +644,9 @@ if __name__ == "__main__":
     )
     args = parser.parse_args()
 
-    if args.slack_channel:
-        with slack_notifications(slack_token, args.slack_channel):
-            main(args)
-    else:
-        main(args)
+    # if args.slack_channel:
+    #     with slack_notifications(slack_token, args.slack_channel):
+    #         main(args)
+    # else:
+    #     main(args)
+    main(args)
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -64,20 +64,21 @@ def main(args):
             logger.info("Creating transcript reference resources...")
             create_transcript_ref(build="GRCh38", overwrite=args.overwrite)
 
-        if args.command == "prep-filtered-context":
-            hl.init(
-                log="/RMC_pre_process.log",
-                tmp_dir=TEMP_PATH_WITH_FAST_DEL,
-                quiet=args.quiet,
-            )
-            hl.default_reference("GRCh38")
-            logger.info("Creating filtered context HT...")
-            n_partitions = 10000
-            create_filtered_context_ht(
-                canonical_only=args.filter_to_canonical,
-                n_partitions=args.n_partitions if args.n_partitions else n_partitions,
-                overwrite=args.overwrite_temp,
-            )
+        # TODO: This is not needed for now since we directly read apply_models HT from constraint as our filtered context
+        # if args.command == "prep-filtered-context":
+        #     hl.init(
+        #         log="/RMC_pre_process.log",
+        #         tmp_dir=TEMP_PATH_WITH_FAST_DEL,
+        #         quiet=args.quiet,
+        #     )
+        #     hl.default_reference("GRCh38")
+        #     logger.info("Creating filtered context HT...")
+        #     n_partitions = 10000
+        #     create_filtered_context_ht(
+        #         canonical_only=args.filter_to_canonical,
+        #         n_partitions=args.n_partitions if args.n_partitions else n_partitions,
+        #         overwrite=args.overwrite_temp,
+        #     )
 
         if args.command == "prep-constraint":
             hl.init(
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -25,7 +25,6 @@ from rmc.resources.rmc import (
     simul_search_round_bucket_path,
     single_search_round_ht_path,
 )
-
 from rmc.utils.constraint import (
     annotate_max_chisq_per_section,
     check_break_search_round_nums,
--- a/rmc/pipeline/regional_constraint.py
+++ b/rmc/pipeline/regional_constraint.py
@@ -27,7 +27,6 @@ from rmc.resources.rmc import (
     single_search_round_ht_path,
 )
 
-# from rmc.slack_creds import slack_token
 from rmc.utils.constraint import (
     annotate_max_chisq_per_section,
     check_break_search_round_nums,
@@ -644,9 +643,4 @@ if __name__ == "__main__":
     )
     args = parser.parse_args()
 
-    # if args.slack_channel:
-    #     with slack_notifications(slack_token, args.slack_channel):
-    #         main(args)
-    # else:
-    #     main(args)
     main(args)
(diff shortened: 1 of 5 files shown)


rmc/pipeline/regional_constraint.py
```diff
@@ -26,7 +26,7 @@
     simul_search_round_bucket_path,
     single_search_round_ht_path,
 )
-from rmc.slack_creds import slack_token
+# from rmc.slack_creds import slack_token
```
ch-kr, line 29: quick note that if you remove this, you'll need to update the code below as well
```
rmc/pipeline/regional_constraint.py:648:33: E0602: Undefined variable 'slack_token' (undefined-variable)
```

rmc/pipeline/regional_constraint.py
```diff
@@ -64,20 +64,21 @@ def main(args):
...
-            )
-            hl.default_reference("GRCh38")
-            logger.info("Creating filtered context HT...")
-            n_partitions = 10000
-            create_filtered_context_ht(
-                canonical_only=args.filter_to_canonical,
-                n_partitions=args.n_partitions if args.n_partitions else n_partitions,
-                overwrite=args.overwrite_temp,
-            )
+        # TODO: This is not needed for now since we directly read apply_models HT from constraint as our filtered context
```
ch-kr, line 67:
```suggestion
        # NOTE: This code block is no longer needed since the expected values are now calculated per-variant upstream
```
also, rather than commenting out this code, it's likely cleaner to just leave it for now (i.e., add this note, but don't change the code otherwise). the commented out code seems to be throwing a variety of pylint errors

rmc/utils/constraint.py
```diff
@@ -509,7 +511,16 @@ def create_constraint_prep_ht(
     :param overwrite: Whether to overwrite Table. Default is True.
     :return: None; writes Table to path.
     """
-    ht = filtered_context.ht()
+    #TODO: Make a resource path for this table.
+    # ht = filtered_context.ht()
+    ht = hl.read_table(
```
ch-kr, line 516: can you ask Julia for the resource name/path of this table so that this path isn't hardcoded? this is to future-proof this code section for future gnomAD releases

rmc/utils/constraint.py
```diff
@@ -509,7 +511,16 @@ def create_constraint_prep_ht(
...
+        "gs://gnomad/v4.1/constraint_an_coverage_corrected/apply_models/transcript_consequences/gnomad.v4.1.per_variant_expected.coverage_corrected.ht"
+        )
+    # Annotate obs and exp columns to be used downstream
+    ht = ht.annotate(
+        observed=ht.calibrate_mu.observed_variants[0],
```
ch-kr, line 521: this is a naive question since I haven't thoroughly dug through Julia's code -- is this the correct field? this reads like these are the observed variants once filtered to high coverage sites only (and we want obs across all sites). if this table doesn't have the correct field, then we may want to add `ht = add_obs_annotation(ht)` here

rmc/utils/constraint.py
ch-kr, line 1: same note about the commented out code here

ch-kr (changes requested):
thanks for adding the code changes! can you remove the `#` commenting out code and run black?
image

rmc/pipeline/regional_constraint.py
Ruchit10: Yeah the reason I removed it is because my new computer is not configured to this like the old one and also I felt its redundant since I'm monitoring the job from the browser directly and not really using this anymore but let me know if we really need to keep this

rmc/utils/constraint.py
Ruchit10: Good point. Will ask her

Ruchit10: I've asked her. Until we have it, is it okay to hard-code this for now? (I've put a note there as a reminder)

rmc/utils/constraint.py
Ruchit10: Done

rmc/resources/rmc.py
```diff
@@ -24,14 +24,14 @@
 from rmc.resources.reference_data import FOLD_K
 from rmc.resources.resource_utils import CURRENT_GNOMAD_VERSION
 
-FREEZES = [1]
+FREEZES = [1, 2]
 """
 RMC/MPC data versions computed with current gnomAD version.
 
 gnomAD v2.1.1 freezes: [1, 2, 3, 4, 5, 6, 7].
```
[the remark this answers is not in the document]
Ruchit10: I've put a note but maybe it's too wordy. Can you help make it more concise?

rmc/pipeline/regional_constraint.py
```diff
@@ -642,8 +644,9 @@ def main(args):
...
-    if args.slack_channel:
-        with slack_notifications(slack_token, args.slack_channel):
-            main(args)
-    else:
-        main(args)
+    # if args.slack_channel:
+    #     with slack_notifications(slack_token, args.slack_channel):
+    #         main(args)
+    # else:
+    #     main(args)
```
Ruchit10, line 651: Keeping it commented for now since I'm still very confused as to removal of this lol..

rmc/utils/constraint.py
```diff
@@ -509,7 +522,15 @@ def create_constraint_prep_ht(
     :param overwrite: Whether to overwrite Table. Default is True.
     :return: None; writes Table to path.
     """
-    ht = filtered_context.ht()
+    # TODO: Make a resource path for this table.
+    # ht = filtered_context.ht()
+    ht = get_per_variant_expected_dataset(
+        directory_post_fix="an_coverage_corrected", path_post_fix="coverage_corrected"
+    ).ht()
+    # Annotate obs and exp columns to be used downstream
```
ch-kr, line 530: can you document here that you're grabbing index 0 because this is an array?

rmc/resources/rmc.py
ch-kr, line 1: the workflow check flags that this file would be reformatted by black -- are you running the same version that is listed in the requirements?

ch-kr (commented):
a few more documentation suggestions

rmc/resources/rmc.py
Ruchit10: Good point. I'll need to check the version

rmc/utils/constraint.py
Ruchit10: Done

rmc/utils/constraint.py
```diff
@@ -347,8 +354,12 @@ def create_possible_hts(
     # Annotate HT globals with models
     possible_ht = possible_ht.annotate_globals(
         # TODO: Uncomment lines below when chrX/chrY, coverage models are ready
-        # plateau_x_models=get_models(model_type="plateau", genomic_region="chrx_non_par").he(),
-        # plateau_y_models=get_models(model_type="plateau", genomic_region="chry_non_par").he(),
+        plateau_x_models=get_models(
```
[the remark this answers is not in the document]
Ruchit10: Done

rmc/utils/constraint.py
```diff
@@ -224,9 +230,9 @@ def calculate_exp_from_mu(
 
     # TODO: uncomment when the other models exist
     # Used this for testing:
-    # plateau_model = hl.experimental.read_expression(
-    #    "gs://gnomad/v4.1/constraint_an/models/gnomad.v4.1.plateau.autosome_par.he"
-    # )
+    plateau_model = hl.experimental.read_expression(
+        "gs://gnomad/v4.1/constraint_an/models/gnomad.v4.1.plateau.autosome_par.he"
+    )
```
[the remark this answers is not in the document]
Ruchit10: Done

rmc/resources/rmc.py
Ruchit10: I was using 24.10.0, re-installed to 23.7.0

ch-kr (commented):
only one minor comment

rmc/utils/constraint.py
```diff
@@ -509,7 +520,16 @@ def create_constraint_prep_ht(
     :param overwrite: Whether to overwrite Table. Default is True.
     :return: None; writes Table to path.
     """
-    ht = filtered_context.ht()
+    # TODO: Make a resource path for this table.
```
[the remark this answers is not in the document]
Ruchit10: Makes sense. I'll delete this TODO then

rmc/utils/constraint.py
```diff
@@ -509,7 +520,15 @@ def create_constraint_prep_ht(
     :param overwrite: Whether to overwrite Table. Default is True.
     :return: None; writes Table to path.
     """
-    ht = filtered_context.ht()
+    # NOTE: Observed counts upstream now includes variants with AF <= 0.001 instead of AF < 0.001.
+    # ht = filtered_context.ht()
```
ch-kr, line 524:
```suggestion
```

rmc/utils/constraint.py
```diff
@@ -509,7 +520,15 @@ def create_constraint_prep_ht(
...
+    ht = get_per_variant_expected_dataset(
+        directory_post_fix="an_coverage_corrected", path_post_fix="coverage_corrected"
```
ch-kr, line 526: maybe these should be arguments to the function with defaults rather than hard-coded (in case the naming scheme changes in the future)

rmc/utils/constraint.py
```diff
@@ -509,7 +520,15 @@ def create_constraint_prep_ht(
...
+    ).ht()
+    # Annotate obs and exp counts as the zero index (pop global) since they are arrays
```
ch-kr, line 528:
```suggestion
    # Obs and exp counts are arrays to account for different frequency strata
    # Use zeroth index (gnomAD-wide ["global"] frequencies)
```
maybe this (0) should also be an argument? something like
```
variant_idx: int = 0,
```

rmc/utils/generic.py
```diff
@@ -506,7 +506,7 @@ def keep_criteria(
     :return: Boolean expression used to filter variants.
     """
     af_filter_expr = (
-        (af_expr < af_threshold) if filter_to_rare else (af_expr >= af_threshold)
+        (af_expr <= af_threshold) if filter_to_rare else (af_expr > af_threshold)
```
ch-kr, line 509: minor note to update the docstring above for the behavior of `filter_to_rare`

ch-kr (commented):
a few nitpicks, then good to go!

rmc/utils/constraint.py
Ruchit10: Yup. Done

rmc/utils/constraint.py
Ruchit10: Can be but I don't think this index will change since 0 always represented global intuitively?

rmc/utils/generic.py
Ruchit10: Done

rmc/utils/constraint.py
```diff
@@ -507,9 +523,21 @@ def create_constraint_prep_ht(
     :param n_partitions: Number of desired partitions for the Table. Default is 15000.
     :param csq: Desired consequences. Default is {`MISSENSE`}. Must be specified if filter is True.
     :param overwrite: Whether to overwrite Table. Default is True.
+    :param directory_post_fix: Specify directory suffix for reading per-variant expected dataset. Default is "an_coverage_corrected".
+    :param path_post_fix: Specify file suffix to use for reading per-variant expected dataset. Default is "coverage_corrected".
```
ch-kr, line 527:
```suggestion
    :param directory_post_fix: Directory suffix for reading per-variant expected dataset. Default is "an_coverage_corrected".
    :param path_post_fix: File suffix to use for reading per-variant expected dataset. Default is "coverage_corrected".
```

rmc/utils/constraint.py
```diff
@@ -507,9 +523,21 @@ def create_constraint_prep_ht(
...
+    :param variant_idx: Index of observed and expected arrays to use. Default is 0.
```
ch-kr, line 528:
```suggestion
    :param variant_idx: Index of observed and expected arrays to use. Default is 0 (corresponds to counts calculated on gnomAD-wide / "global" frequencies).
```

rmc/utils/constraint.py
```diff
@@ -507,9 +523,21 @@ def create_constraint_prep_ht(
...
     :return: None; writes Table to path.
     """
-    ht = filtered_context.ht()
+    # NOTE: Observed counts upstream now includes variants with AF <= 0.001 instead of AF < 0.001.
+    ht = get_per_variant_expected_dataset(
+        directory_post_fix=directory_post_fix, path_post_fix=path_post_fix
+    ).ht()
+    # Obs and exp counts are arrays to account for different frequency strata
+    # Use zeroth index (gnomAD-wide ["global"] frequencies)
```
ch-kr, line 536:
```suggestion
```

ch-kr (approved):
a couple documentation updates then LGTM

PR #314, opened 2025-05-20
Runtime fixes to RMC break search and generating the regions 

Ruchit10:
Major change:
1. Added step to filter the upstream constraint HT to canonical ensembl transcripts during `prep-constraint` step

Minor runtime fixes:
1. Removed slack notifications part throughout the pipeline.
2. Added feature to specify a custom google project in the required steps of the pipeline.
3. Minor variable assignment shift and adding dummy variable in `run_batches_dataproc.py` to avoid pre-commit errors

### DIFF
--- a/rmc/utils/constraint.py
+++ b/rmc/utils/constraint.py
@@ -538,6 +538,9 @@ def create_constraint_prep_ht(
     # NOTE: All MANE Select transcripts in GENCODE v39 are canonical, but not all canonical transcripts are MANE Select
     logger.info("Filtering to canonical Ensembl transcripts...")
     ht = ht.filter(ht.canonical & ht.transcript.startswith("ENST"))
+    # Count number of transcripts after filtering
+    n_transcripts = len(ht.aggregate(hl.agg.collect_as_set(ht.transcript)))
+    logger.info("Found %d transcripts after filtering...", n_transcripts)
     # Obs and exp counts are arrays to account for different frequency strata
     ht = ht.annotate(
         observed=ht.calibrate_mu.observed_variants[variant_idx],
@@ -1155,6 +1158,8 @@ def check_break_search_round_nums(
         Assumes there is a folder for each search round run, regardless of whether there were breaks discovered
 
     :param freeze: RMC freeze number. Default is CURRENT_FREEZE.
+    :param google_project: Google project to use to read data from requester-pays buckets.
+        Default is 'broad-mpg-gnomad'.
     :return: Sorted list of round numbers.
     """
     # Get sorted round numbers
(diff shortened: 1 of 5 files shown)


rmc/utils/constraint.py
```diff
@@ -534,6 +534,9 @@ def create_constraint_prep_ht(
     ht = get_per_variant_expected_dataset(
         directory_post_fix=directory_post_fix, path_post_fix=path_post_fix
     ).ht()
+    # filter to Canonical Ensembl transcripts
```
ch-kr:
```suggestion
    # Filter to Canonical Ensembl transcripts
    # NOTE: All MANE Select transcripts in GENCODE v39 are canonical, but not all canonical transcripts are MANE Select
```
(black may reformat this if it's too long)

rmc/utils/constraint.py
```diff
@@ -1137,7 +1140,10 @@ def get_break_search_round_nums(
     return sorted(round_nums)
 
 
-def check_break_search_round_nums(freeze: int = CURRENT_FREEZE) -> List[int]:
+def check_break_search_round_nums(
+    freeze: int = CURRENT_FREEZE,
+    google_project: str = "broad-mpg-gnomad",
```
ch-kr, line 1149: thanks for adding! this also needs to be added to the function docstring

rmc/utils/constraint.py
```diff
@@ -534,6 +534,9 @@ def create_constraint_prep_ht(
     ht = get_per_variant_expected_dataset(
         directory_post_fix=directory_post_fix, path_post_fix=path_post_fix
     ).ht()
+    # filter to Canonical Ensembl transcripts
+    logger.info("Filtering to canonical Ensembl transcripts...")
+    ht = ht.filter(ht.canonical & ht.transcript.startswith("ENST"))
```
ch-kr, line 540: it would add extra time, but can you add an aggregator here to count and print the total number of transcripts after this filter?

ch-kr (commented):
a couple minor things

rmc/utils/constraint.py
Ruchit10: oops. Forgot about this. Will update docstring..

rmc/utils/constraint.py
Ruchit10: sounds good, will do.

rmc/utils/constraint.py
Ruchit10: Done

rmc/utils/constraint.py
Ruchit10: Done

ch-kr (approved):
LGTM!

PR #315, opened 2025-07-10
Switch to new constraint resource path for building browser HT and update context with OE to use browser table

Ruchit10:
List of Changes:
1. `get_constraint_transcripts` now uses new upstream constraint table to filter to canonical protein coding ENST 
2. Default resource path for `constraint_ht` is updated to point to new gene constraint table
3. `create_context_with_oe` will use the browser table instead of regions table. This also fixed a bug that the regions OE was still capped to 1 while transcript OE was uncapped.
4. `get_oe_annotation` is modified to explode the regions table into loci followed by a left join to annotate context table with region OE

### DIFF
--- a/rmc/resources/gnomad.py
+++ b/rmc/resources/gnomad.py
@@ -10,7 +10,7 @@ constraint_ht = VersionedTableResource(
             path="gs://gcp-public-data--gnomad/release/2.1.1/constraint/gnomad.v2.1.1.lof_metrics.by_transcript.ht"
         ),
         "4.1": TableResource(
-            path="gs://gcp-public-data--gnomad/release/4.1/constraint/gnomad.v4.1.constraint_metrics.ht"
+            path="gs://gcp-public-data--gnomad/release/4.1/constraint_cov_corrected/gnomad.v4.1.constraint_metrics.ht"
         ),
     },
 )
--- a/rmc/resources/gnomad.py
+++ b/rmc/resources/gnomad.py
@@ -10,9 +10,8 @@ constraint_ht = VersionedTableResource(
         "4.1": TableResource(
             path="gs://gcp-public-data--gnomad/release/4.1/constraint/gnomad.v4.1.constraint_metrics.ht"
         ),
-        # TODO: Update this path when the new public version is released.
         "4.1.1": TableResource(
-            path="gs://gnomad/v4.1/constraint_coverage_corrected/metrics/transcript_consequences/gnomad.v4.1.constraint_metrics.coverage_corrected.ht"
+            path="gs://gcp-public-data--gnomad/release/4.1.1/constraint/gnomad.v4.1.1.constraint_metrics.ht"
         ),
     },
 )
--- a/rmc/resources/gnomad.py
+++ b/rmc/resources/gnomad.py
@@ -21,7 +21,7 @@ Public gnomAD gene constraint Table.
 
 v2.1.1: Observed variants count is annotated as `ht.obs_mis` and expected variants count is annotated as `ht.exp_mis`.
 v4.1: Observed variants count is annotated as `ht.mis.obs` and expected variants count is annotated as `ht.mis.exp`.
-NOTE: default version is manually set to point to most updated resource. When public resource is updated,
-please set the default version to CURRENT_GNOMAD_VERSION and add this import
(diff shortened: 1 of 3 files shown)


rmc/resources/gnomad.py
```diff
@@ -1,17 +1,21 @@
 """Script containing resources for the current gnomAD version."""
 from gnomad.resources.resource_utils import TableResource, VersionedTableResource
 
-from rmc.resources.resource_utils import CURRENT_GNOMAD_VERSION
+# from rmc.resources.resource_utils import CURRENT_GNOMAD_VERSION
```
ch-kr: instead of commenting this line out, can you update `CURRENT_GNOMAD_VERSION` to 4.1.1?

rmc/utils/constraint.py
```diff
@@ -1429,24 +1430,26 @@ def get_oe_annotation(ht: hl.Table, freeze: int) -> hl.Table:
...
-        section_oe=hl.or_missing(
-            hl.len(ht.section_oe) > 0,
-            ht.section_oe[0],
-        ),
-    )
+    rmc_exploded = explode_intervals_to_loci(rmc_ht).key_by("locus", "transcript")
+    ht = ht.annotate(section_oe=rmc_exploded[ht.locus, ht.transcript].regions.oe)
+
     # Annotate with RMC OE where available, otherwise select transcript OE
     return ht.transmute(oe=hl.coalesce(ht.section_oe, ht.transcript_oe))
```
ch-kr, line 1528: this table isn't filtered to remove non-protein-coding transcripts, right? we should probably add that filter

ch-kr (commented):
thank you for adding! just a couple minor comments. it also looks like the code needs to be updated with the current main branch:
image

rmc/utils/constraint.py
ch-kr: @Ruchit10 pointed out that I missed this: https://github.com/broadinstitute/regional_missense_constraint/blob/aa90a3b1927faca7eb0088b7613214d748af8d6f/rmc/utils/constraint.py#L1536 so the table should be filtered

rmc/resources/gnomad.py
```diff
@@ -1,22 +1,27 @@
...
     },
 )
 """
 Public gnomAD gene constraint Table.
 
 v2.1.1: Observed variants count is annotated as `ht.obs_mis` and expected variants count is annotated as `ht.exp_mis`.
 v4.1: Observed variants count is annotated as `ht.mis.obs` and expected variants count is annotated as `ht.mis.exp`.
+NOTE: default version is manually set to point to most updated resource. When public resource is updated,
+please set the default version to CURRENT_GNOMAD_VERSION and add this import
+from rmc.resources.resource_utils import CURRENT_GNOMAD_VERSION
```
ch-kr:
```suggestion
NOTE: Default version is manually set to point to most updated resource. When public resource is updated,
please update `CURRENT_GNOMAD_VERSION`, set the default version to `CURRENT_GNOMAD_VERSION` and add this import:
`from rmc.resources.resource_utils import CURRENT_GNOMAD_VERSION`
```

rmc/utils/constraint.py
```diff
@@ -1429,24 +1430,26 @@ def get_oe_annotation(ht: hl.Table, freeze: int) -> hl.Table:
...
-        .filter(lambda x: x.transcript == ht.transcript)
-        .section_oe
-    )
-    ht = ht.annotate(
-        section_oe=hl.or_missing(
-            hl.len(ht.section_oe) > 0,
-            ht.section_oe[0],
-        ),
-    )
+    rmc_exploded = explode_intervals_to_loci(rmc_ht).key_by("locus", "transcript")
```
ch-kr:
```suggestion
    rmc_exploded = explode_intervals_to_loci(rmc_ht, interval_field="interval", keep_intervals=False).key_by("locus", "transcript")
```
this depends on suggestions to the other PR (and also might not be correctly formatted)

ch-kr (commented):
two minor things (one documentation update, and one update to `explode_intervals_to_loci` based on suggestions in other PR, then LGTM

ch-kr:
merging now in the interest of time

ch-kr (approved):

PR #323, opened 2026-06-24
Update browser HT

ch-kr:
After discussion, update browser release HT to:
- Contain exact missense o/e percentile value (rather than <= value)
- Update globals to add new fields indicating which transcripts without evidence of RMC have gene flags


ch-kr:
closing since these changes were pulled into https://github.com/broadinstitute/regional_missense_constraint/pull/324

PR #324, opened 2026-08-18
Run RMC on the MANE Plus Clinical transcripts (freezes 3 and 4)

ch-kr:
# Run RMC on the MANE Plus Clinical transcripts (freezes 3 and 4)

Extends the v4 RMC pipeline to run on an arbitrary transcript set, uses that to search
the 56 transcripts unique to the MANE Select plus clinical set as **freeze 3**, and
unions freezes 2 and 3 into **freeze 4** for release.

## What's here

**Arbitrary transcript sets.** `create_constraint_prep_ht` takes `keep_transcripts` and
filters to it instead of to canonical transcripts. `create-mane-plus-clinical-he` parses
the public GENCODE v39 GTF to find MANE Plus Clinical transcripts that aren't Ensembl
canonical, and writes them to a HailExpression.

**Released transcript sets.** `add_globals_rmc_browser` scopes its sets to the
transcripts a freeze covers (canonical plus `--extra-transcripts-he`) and raises when a
freeze released non-canonical transcripts without declaring them.

**Coverage stats.** New `create-coverage-stats` step writes median exome AN percent per
RMC region, computed over CDS sites only and keyed on the amino acid adjusted
coordinates the browser displays. Transcripts without evidence of RMC are deliberately
absent: the public constraint Table already carries their coverage as the
`low_exome_coverage` gene flag, which the browser globals release.

**Freeze unions.** `union-freezes` unions named freeze-versioned resources into an
output freeze, checking that globals match and that transcripts are disjoint.
`create-rmc-release --union-freezes-aa` reuses the input freezes' amino acid
annotations rather than repeating the lookup. Note that nothing ties the two freeze
lists together, so passing a strict subset to `--union-freezes-aa` would silently drop
the missing freeze's transcripts; passing a *wrong* freeze is caught downstream, since
its regions find no row in freeze 4's coverage Table.

**Merged** `kc/update_percentiles` (#323): integer missense O/E percentiles and the
no-RMC gene flag globals.

## Running it

```
create-mane-plus-clinical-he
prep-constraint       --freeze 3 --keep-transcripts-he <mane_plus_clinical_transcripts_path>
<break search rounds> --freeze 3
finalize              --freeze 3
create-coverage-stats --freeze 3
union-freezes         --union-freezes 2 3 --output-freeze 4 --union-resources rmc_coverage_stats
create-rmc-release    --freeze 4 --extra-transcripts-he <path> --union-freezes-aa 2 3
```

Start the cluster with **`--no-off-heap-memory`**. Without it, the amino acid lookup in
`create-coverage-stats` fails with `Hail off-heap memory exceeded maximum threshold`; a
highmem machine type, per-job memory properties, and tighter intervals all failed to fix
it.

## Test run

Ran end to end on `n1-standard-8` workers (2 primary + up to 20 preemptible), code as of
`63bd737`.

| Step | Dataproc job | Duration |
|---|---|---|
| `create-mane-plus-clinical-he` | `be45078089ed45768c991e4c17e74c93` | 54 s |
| `prep-constraint --freeze 3` | `23d94ff231a9413c9454fe811cc4a49a` | 12m52s |
| break search, 10 rounds | (per-round jobs) | ~4 h total |
| `create-coverage-stats --freeze 3` | `08d8894b1a4f419caa1df051921ab504` | 73 min |
| `union-freezes 2 3 -> 4` | `945e8975e0a642aaa48cf04542b79cb6` | 2 min |
| `create-rmc-release --freeze 4` | `1ff9392651c6401d92c1cbbe9e5a4ac6` | 38 min |

Freeze 3 converged in 10 rounds (freeze 2 needed 16): 56 transcripts in, 36 with RMC
across 251 regions, 20 without.

Two timing notes for whoever runs this next. `create-coverage-stats` spent 34m04s of its
73 minutes inside `_calculate_new_partitions`, from a `_n_partitions` argument that was
added to fight the off-heap failure and didn't; removing it (`63bd737`) is why the
release step below it takes 38 minutes rather than ~70. The release's own largest stage
is 14m42s of amino acid lookup over 94,181 CDS intervals, which `--union-freezes-aa`
exists to avoid on later releases.

## Freeze 4 verification

```
regions              28,058  = 27,807 (freeze 2) + 251 (freeze 3)
transcripts w/ RMC    6,775
MPC transcripts          56  = 36 with RMC + 20 without, 0 unaccounted
MPC in covered set    56/56
QC-pass transcripts  17,895  = 17,841 canonical + 54 QC-pass MPC
all covered          19,431  = 19,375 canonical + 56 MPC
no-RMC low coverage     304  (413 carry any gene flag)
```

`no_breaks` for freeze 4 holds 13,288 transcripts against 12,656 in the globals. The 632
difference is freeze 2 `no_breaks` transcripts outside the canonical set, from the freeze
2 `constraint_prep` scoping issue. The globals come from the constraint Table and are the
number to quote.

## Not yet executed on a cluster

The freeze 3 and 4 outputs currently in the bucket were produced by earlier commits in
this branch. These landed afterwards and have **never run on Dataproc**:

- `get_rmc_regions_aa_ht` — the shared derivation of adjusted coordinates that both
  `create-coverage-stats` and `create-rmc-release` now depend on
- `union_rmc_browser_regions` and `--union-freezes-aa` — freeze 4's release recomputed
  amino acids instead of reusing them
- the freeze validity and disjointness guards on the union path

Schema-level behaviour of the union was smoke-tested locally against freeze 2's real
browser Table. A single freeze 4 re-release with `--union-freezes-aa 2 3` would exercise
all of the above and produce the timing comparison for the amino acid reuse.

## Decisions worth a look

- **MPC and missense badness stay on freeze 2.** Those steps aren't rerun for freeze 3.
- **`CONSTRAINT_VERSION` pinned to 4.1.** gnomad-constraint's default moved to 4.1.1,
  which has no coverage corrected per-variant expected dataset, so freeze 3 would have
  read a different model than freeze 2.
- **Low coverage threshold applied at release, not stored.** Keeps
  `rmc_coverage_stats_ht` unionable with freeze 2's existing Table.
- **`create_rmc_release_downloads` is unreachable and broken.** Documented rather than
  fixed; reviving it means choosing QC-pass versus all-transcripts for the public no-RMC
  TSV.

## Known issue found while reviewing, fixed on a separate branch

`process_section_group`'s over-threshold branch skips one candidate first breakpoint per
`split_list_len` block (`rmc/utils/simultaneous_breaks.py:434`). Confirmed by rerunning
the real freeze 3 round 1 search and reading back the indices it recorded: for section
`ENST00000413689_38550324_38633307` (5,026 sites), `i` values 499, 999, … 4999 are never
evaluated, while `j` is contiguous.

Pre-existing, and it affects every section spanning more than one 500-index block —
`run_batches_dataproc.py` searches all sections block by block regardless of the 5,000
site threshold, which only names output files:

| | Section-searches | Affected (`max_idx >= 500`) | Skipped `i` |
|---|---|---|---|
| Freeze 2 (rounds 1-16) | 48,280 | 24,553 | 57,110 |
| Freeze 3 (rounds 1-10) | - | 141 | - |

A result only changes if a section's optimum sat on a skipped index. The 437 largest
sections have been rerun with the fix and are unchanged; a run over the rest is in
progress. The fix is on `kc/fix_two_breaks_index_gap` (#325), kept out of this PR because
it's pre-existing and its scope is a results question.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rmc/resources/rmc.py
```diff
@@ -26,16 +26,30 @@
...
+expected variant counts and the gene constraint HT are both v4.1.1), but resource
+paths still use `CURRENT_GNOMAD_VERSION` ("4.1") because RMC on v4 started on v4.1.
+The exception is the all sites AN HT used for coverage stats, which is only released
+for v4.1.
+
+NOTE: Missense badness and MPC were not regenerated for freezes 3 and 4 -- they remain
+on freeze 2, which is why `context_with_oe` was not unioned into freeze 4.
 """
 
 CURRENT_FREEZE = 2
```
Ruchit10: Can we add a note here explaining why current freeze is deliberately kept 2 while the release data freeze would be 4+?

rmc/resources/rmc.py
```diff
@@ -535,6 +566,33 @@ def no_breaks_he_path(
...
+UNIONABLE_RESOURCES = {
+    "rmc_results": rmc_results,
+    "rmc_coverage_stats": rmc_coverage_stats_ht,
+}
+"""
+Freeze-versioned Tables that can be unioned across freezes.
+
+Used by `union_freeze_resources` to look up resources by name.
+
+NOTE: `context_with_oe` and `context_with_oe_dedup` aren't included because they are
```
Ruchit10, line 588: Should we include a bigger note somewhere in the main script as a TODO?

rmc/utils/constraint.py
```diff
@@ -2150,38 +2343,381 @@ def annot_rmc_with_percentile(ht: hl.Table, oe_field: str) -> hl.Table:
...
+            max_oe,
+        )
+
+    # Find the smallest index whose percentile value is larger than or equal to the provided
+    # OE. The percentile is that index + 1 (percentiles start at 1).
+    match = hl.enumerate(hl.literal(mis_oe_pcts)).find(lambda t: t[1] >= oe)
+    return ht.annotate(mis_oe_percentile=hl.if_else(oe > max_oe, 100, match[0] + 1))
+
+
+def get_exomes_an_percent_expr(
```
Ruchit10, line 2382: Why not use this function instead? Is it because of non-main constraint branch? https://github.com/broadinstitute/gnomad-constraint/blob/0470d69a3941f4d5df55aec33d17424ac9d1fb92/gnomad_constraint/utils/constraint.py#L374

rmc/utils/constraint.py
```diff
@@ -2150,38 +2343,381 @@ def annot_rmc_with_percentile(ht: hl.Table, oe_field: str) -> hl.Table:
...
+    for name in resource_names:
+        logger.info("Unioning %s across freezes %s...", name, freezes)
+        if name == NO_BREAKS_RESOURCE:
+            transcript_sets = [
+                set(hl.eval(hl.experimental.read_expression(no_breaks_he_path(freeze))))
+                for freeze in freezes
+            ]
+            _check_transcripts_disjoint(transcript_sets, name)
+            hl.experimental.write_expression(
+                hl.literal(set().union(*transcript_sets)),
```
Ruchit10, line 2593: This has not been an issue since neither of the `transcript_sets` are empty but do you think we should add a guardrail here for any possible empty set?

rmc/utils/data_loading.py
```diff
@@ -36,6 +42,126 @@
 ####################################################################################
 ## Transcript utils
 ####################################################################################
+TRANSCRIPT_ID_REGEX = re.compile(r'transcript_id "([^".]+)')
```
Ruchit10, line 45: Should we guard this with a scenario of a transcript feature missing `transcript_id`? May happen with a malformed GTF line. Adding a `None` guard to `get_gencode_transcript_tags` should fix it?

rmc/utils/constraint.py
```diff
@@ -2150,38 +2343,381 @@ def annot_rmc_with_percentile(ht: hl.Table, oe_field: str) -> hl.Table:
...
+        interval=hl.interval(
+            rmc_ht.start_coordinate,
+            rmc_ht.stop_coordinate,
+            includes_start=True,
+            includes_end=True,
+        ),
+        transcript=rmc_ht.transcript,
+    ).select()
+
+    regions_ht = rmc_ht
```
Ruchit10: Is this alias necessary?

rmc/utils/constraint.py
```diff
@@ -2150,38 +2343,381 @@ def annot_rmc_with_percentile(ht: hl.Table, oe_field: str) -> hl.Table:
...
+    requires the amino acid lookup over the VEP context Table, the most expensive step
+    in the pipeline.
+
+    `create_rmc_coverage_stats` needs these coordinates to key coverage on the regions
+    the browser displays, and `format_rmc_browser_ht` needs them plus the amino acids,
+    so both read this checkpoint rather than running the lookup once each.
+
+    .. warning::
+
+        The checkpoint is keyed on the freeze alone, so rerunning `finalize` leaves it
```
Ruchit10, line 2630: Just thinking whether it should be keyed to something else or some other guardrails for an accidental `overwrite_temp=False`?

rmc/utils/constraint.py
```diff
@@ -2150,38 +2343,381 @@ def annot_rmc_with_percentile(ht: hl.Table, oe_field: str) -> hl.Table:
...
+                section_chisq=ht.regions.chisq,
+            )
+        )
+    transcript_sets = [ht.aggregate(hl.agg.collect_as_set(ht.transcript)) for ht in hts]
+    n_transcripts = sum(len(t) for t in transcript_sets)
+    if len(set().union(*transcript_sets)) != n_transcripts:
+        raise DataException(
+            f"Transcripts overlap across browser Tables for freezes {freezes}!"
+        )
+    return hts[0].union(*hts[1:])
```
Ruchit10: Should we add a check if the union transcripts cover every transcript in `rmc_results.versions[output_freeze]` or is that unnecessary?

rmc/utils/constraint.py
```diff
@@ -2229,40 +2768,41 @@ def format_rmc_browser_ht(
     :param overwrite_temp: Whether to overwrite temporary data.
```
Ruchit10, line 2781: Very minor but should we add note here that this is not used when `union_freezes` is set?

Ruchit10 (changes requested):
Mostly minor things which shouldn't affect the run but good to address. Hope I didn't miss any major code breaks

rmc/resources/rmc.py
ch-kr: these code changes were added before the bug (fixed in 325) was found. we should update this number to whichever number we're going to use for the freeze 2 rerun (3?). also, as I noted in a scrum, we'll actually need more than the number of freezes introduced in this PR to fully account for the freeze 2 rerun, the final MANE Plus Clinical run, and the union of the rerun + MANE Plus Clinical; I haven't added this yet to avoid updating the code while you were actively reviewing

rmc/resources/rmc.py
ch-kr: I don't think this should be a TODO unless we want to generate these resources with MANE Plus Clinical and not just the set of canonical transcripts

rmc/utils/constraint.py
ch-kr: I didn't realize we had that function, and yes, I'd prefer not to pull from a non-reviewed branch for anything that we are releasing on the gnomAD browser

rmc/utils/constraint.py
ch-kr: can you explain this more?

rmc/utils/data_loading.py
ch-kr: we could, but I'm not sure it's worth it; we shouldn't need to run this code more than once (unless we need it in the future), and it's already been run on this GENCODE version without any errors. I can add it if you feel strongly about it, though

rmc/utils/constraint.py
ch-kr: thanks for flagging! no, this probably was added in an intermediate version and doesn't seem necessary anymore

rmc/utils/constraint.py
ch-kr: I can add this if you feel strongly about it, but I think we'd want to avoid passing overwrite_temp in most cases where we rerun this `finalize` step for the same freeze within a 2 week window

rmc/resources/rmc.py
ch-kr: I just updated the freezes and the description. question for you: since we are rerunning freeze 2, we could ignore some of the structure added in this PR and run all transcripts (canonical + MANE Plus Clinical) in the next freeze (3). however, that would probably require some code tweaks, and we could strip the union code from the repo, which you've already reviewed.

it isn't the cleanest, but I'm OK with running freeze 3 (freeze 2 rerun) and freeze 4 (MANE Plus Clinical addition) simultaneously and then unioning them to create freeze 5, since that might be the fastest path to getting results ready. let me know what you prefer

rmc/utils/constraint.py
Ruchit10: Should we add a note above in that case so we can remove this once the constraint PR is merged and add a function import?

rmc/utils/constraint.py
Ruchit10: So the case when none of the transcripts have RMC (or conversely when all transcripts show RMC then one of the sets would be empty but we know that won't happen with the current MANE + clinical set so this won't error (this was just a precaution if someone decided to run on a custom set of transcripts

rmc/utils/constraint.py
Ruchit10: I think this warning note should suffice but not sure if the note should be extended further to emphasize. Because a rerun on same freeze would read the old table and there would be no way to know from the logs alone. We could also add a log message like "Reading from the old table. If this is a rerun on same freeze read the warning message" or something similar.

rmc/utils/data_loading.py
Ruchit10: I think we can leave this as is. You're right that this was run on current GENCODE without any issues and on the off chance that this is run on a broken GTF or something it should error out leading to a fix in the input which serves the purpose.

Ruchit10 (changes requested):
Thanks for addressing my previous comments. Just a few more minor ones

rmc/utils/constraint.py
ch-kr: 🤔 this shouldn't error, since there should be transcripts in both bins (with and without evidence of RMC), but I can add a check if you feel strongly about this

rmc/utils/constraint.py
ch-kr: there is a warning below (`logger.info("Reading existing amino acid annotated RMC regions...")`); I could make this a logger.warning?

rmc/resources/rmc.py
Ruchit10: I think the current freeze structure looks good for the rerun and is the quickest. We can implement the cleaner structure later if and when needed

rmc/utils/constraint.py
Ruchit10: I think it looks good for the current clinical set. I was thinking more of a (custom) set for freeze 4 (instead of current clinical set) where no transcripts show RMC but arguably that's not going to be a part of release so doesn't matter

rmc/utils/constraint.py
Ruchit10: Ah I missed that line. That should be fine then

Ruchit10 (approved):
LGTM

PR #325, opened 2026-08-18
Search every candidate first breakpoint in split sections

ch-kr:
# Search every candidate first breakpoint in split sections

Stacked on #324 — base is `kc/mane_plus_clinical_freeze`, so this diff is just the fix.

## The bug

Sections above the 5,000-site threshold are split into blocks of `split_list_len`
(default 500), and each row of the block grid got:

```python
i_max_idx = hl.min(ht.i + split_list_len, ht.max_idx) - 1   # -1 applied per block
j_max_idx = hl.min(ht.j + split_list_len, ht.max_idx)       # no -1
```

Both are consumed as **exclusive** `hl.range` ends. The `-1` is meant to exclude only the
section's last index — `i = max_idx` can't form a window, since a window needs `j > i` and
`calculate_window_chisq` is missing at `j == max_idx`. That's implemented correctly in the
non-split branch (`i_max_idx = ht.max_idx - 1`). In the split branch the same `-1` lands on
every block, so each block stops one index short while the next starts at the following
multiple of 500.

Worked example, `max_idx = 12`, `split_list_len = 5`:

| Row | `i` searched | `j` searched |
|---|---|---|
| `i_start=0, j_start=0` | 0-3 | 1-4 |
| `i_start=0, j_start=5` | 0-3 | 5-9 |
| `i_start=5, j_start=5` | 5-8 | 6-9 |

`i = 4` is never searched. In the first row that's correct — no `j > 4` is available there.
But the second row searches `j = 5…9`, where `(4,5) … (4,9)` are all valid windows that go
unscored. `j` has no gap, because it has no `-1`.

## Confirmed on production data

Reran the real freeze 3 round 1 search (`prepare_transcripts.py` output, `over_threshold=True`,
production defaults) and read back the `i`/`i_max_idx` values the search itself recorded:

```
section ENST00000413689_38550324_38633307, max_idx=5026, 66 block rows
i searched: 5014 of 5024 valid
i never searched: 499, 999, 1499, 1999, 2499, 2999, 3499, 3999, 4499, 4999
j searched: contiguous 0..5025
```

**Every section searched for two breaks is affected, not just the ones above the 5,000
site threshold.** `run_batches_dataproc.py` computes an `over_threshold` variable from its
CLI flag but uses it only to name output files, and passes the literal `over_threshold=True`
to `process_section_group`. A section escapes the gap only when it fits in a single block,
i.e. fewer than `split_list_len` sites.

Checked against the unfixed code on real sections: `max_idx=291` (one block) loses nothing,
`max_idx=698` (three block rows) loses exactly `i=499`.

| | Section-searches | Affected (`max_idx >= 500`) | Skipped `i` |
|---|---|---|---|
| Freeze 2 (rounds 1-16) | 48,280 | 24,553 | 57,110 |
| Freeze 3 (rounds 1-10) | - | 141 | - |

## The fix

```python
i_max_idx = hl.min(ht.i + split_list_len, ht.j_max_idx - 1, ht.max_idx - 1)
```

`i` now stops below both the section's last index and the row's last `j`. Blocks are
contiguous, and rows still skip `i` values that have no `j` to pair with, so the diagonal
block doesn't gain useless evaluations. `j_max_idx` moves to its own `annotate` call because
a field can't be referenced in the call that creates it.

## Validation: 10 released sections change

Every affected section on both freezes was rerun with the fix into a scratch prefix and
diffed against the original `final_results/merged.ht` per round.

```
checked=24694  new=0  changed=10  lost=0
```

| Transcript | Freeze/round | Recovered `i` | chi-square before -> after |
|---|---|---|---|
| ENST00000246186 | 2 / 1 | 499 | 16.91 -> 17.37 |
| ENST00000253693 | 2 / 1 | 499 | 14.77 -> 14.84 |
| ENST00000360835 | 2 / 1 | 1499 | 19.79 -> 20.09 |
| ENST00000381668 | 2 / 1 | 1999 | 17.73 -> 17.84 |
| ENST00000559447 | 2 / 1 | 999 | 17.84 -> 18.09 |
| ENST00000286835 | 2 / 2 | 499 | 21.13 -> 21.28 |
| ENST00000325324 | 2 / 2 | 999 | 14.95 -> **16.61** |
| ENST00000396946 | 2 / 2 | 1999 | 14.44 -> 14.77 |
| ENST00000303924 | 2 / 3 | 499 | 19.53 -> 20.22 |
| ENST00000280772 | 2 / 4 | 499 | 16.62 -> 16.63 |

Each change is attributable to the fix rather than to this branch's refactor:

- the new first breakpoint lands on a block boundary the old code could not reach
- chi-square rose in all 10, which a maximum over a superset of candidates must do
- no section lost a break, and no section gained one it did not have
- a control run of this branch's code with the *old* index formula reproduced production
  exactly (`mismatches=0`), so removing the unsplit branch changed no behaviour

Seven changes move a breakpoint by 1-3 bp. `ENST00000325324` moves both breakpoints
(62485424 -> 62485423 and 62487791 -> 62487774) and gains 1.66 chi-square; it is the one
change that looks material rather than a boundary nudge.

**This is a lower bound on divergence.** Each round was rerun against production's inputs
for that round. End to end, a changed breakpoint in round 1 changes which sections enter
round 2, so the five round-1 changes would cascade. Measuring the real downstream effect
means rerunning freeze 2 from round 1 with the fix — a decision about whether 10 changed
transcripts warrant a re-release, not something this PR settles.

Freeze 3 is unaffected: all 141 of its affected sections are identical.

The run took 2.5 hours total across 24,694 sections. Note that a first attempt on 200
preemptible secondaries died on `FetchFailedException` when reclaimed workers took their
shuffle blocks with them; it completed on 40 non-preemptible workers.

## Also in here

**The unsplit search path is gone.** `process_section_group` had an `over_threshold`
parameter selecting between block-by-block and whole-section search, but the only caller
always passed `True`, so the unsplit branch never ran. It also made the size threshold read
as a code path switch, which is why sections below it looked unaffected when they were not.
A section smaller than `split_list_len` now gets a single block whose bounds match what the
removed branch set.

**`rmc/pipeline/two_breaks/run_batches.py` is removed**, the Hail Batch runner. It carried its own copy of
the search functions because Batch PythonJobs can't import this repo, and the copies had
drifted — its `over_threshold` path aggregated `hl.agg.max(ht.max_chisq)`, a field nothing
produces, so it would raise. It only ever ran for gnomAD v2; the README now points at the
`rmc_gnomad_v2` branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rmc/pipeline/two_breaks/run_batches_dataproc.py
```diff
@@ -231,11 +232,6 @@ def main(args):
         help="Search for simultaneous breaks in sections that are under length cutoff.",
         action="store_true",
     )
-    parser.add_argument(
```
Ruchit10, line 234: Any specific reason for removing this block?

ch-kr: reading the temporary file wasn't actually saving too much time in this case and to avoid accidentally reading the incorrect file if partway through a run

rmc/utils/simultaneous_breaks.py
```diff
@@ -474,16 +460,15 @@ def process_section_group(
...
-    if over_threshold:
-        ht = ht.checkpoint(
-            f"{TEMP_PATH_WITH_FAST_DEL}/{section_group[0]}.ht", overwrite=True
-        )
-        # If any rows had a significant breakpoint,
-        # find the one "best" breakpoint (breakpoint with largest chi square value)
-        if ht.count() > 0:
-            ht = annotate_max_chisq_per_section(ht, freeze)
-            ht = ht.filter(ht.chisq == ht.section_max_chisq)
+    # A split section has one row per block, so reduce to its best break
```
Ruchit10:
```suggestion
    # Checkpoint here to avoid ht.count() on the next step to re-trigger the search expression
```

rmc/utils/simultaneous_breaks.py
```diff
@@ -474,16 +460,15 @@ def process_section_group(
...
+    ht = ht.checkpoint(
+        f"{TEMP_PATH_WITH_FAST_DEL}/{section_group[0]}.ht", overwrite=True
+    )
+    # If any rows had a significant breakpoint,
```
Ruchit10, line 469:
```suggestion
    # A split section has one row per block, so reduce to its best break
    # If any rows had a significant breakpoint,
```

rmc/utils/simultaneous_breaks.py
ch-kr: actually, is a comment necessary here at all? should I just remove the comment?

rmc/utils/simultaneous_breaks.py
ch-kr: this comment seems redundant with "find the one 'best' breakpoint" below

rmc/utils/simultaneous_breaks.py
Ruchit10: I'm okay with removing it but it was good to know that this checkpoint was actually necessary for performance

rmc/utils/simultaneous_breaks.py
Ruchit10: Yes seems redundant with the one below. Fine to remove

rmc/utils/simultaneous_breaks.py
ch-kr: added a comment here

rmc/utils/constraint.py
Ruchit10, line 913: This is fine if run in `run_all_sections` but if both `sections_over_threshold` and `sections_under_threshold` are run concurrently (two parallel jobs?) then there will be a conflict. Should we add conditional naming or add section group here?

ch-kr: over and under threshold are deprecated (leftover behavior from the older pipeline structure), so that shouldn't happen. I could remove support for those options, but it would bloat this PR. what do you think?

rmc/pipeline/two_breaks/run_batches_dataproc.py
```diff
@@ -47,6 +48,8 @@ def main(args):
             chisq_threshold = hl.eval(hl.qchisqtail(args.p_value, 2))
 
         # Adding dummy values to variables to avoid pre-commit errors
+        # NOTE: `over_threshold` only picks the output file name; every section is
+        # searched block by block regardless of size
         sections_to_run = list()
         over_threshold = None
         if args.run_all_sections:
```
Ruchit10, line 55: This will probably raise in later rounds where one of the two expressions are missing due to empty sections, correct?

ch-kr: correct https://github.com/broadinstitute/rmc_production/issues/215

rmc/utils/constraint.py
Ruchit10: oh okay I got confused with the docstring  somewhere mentioning they can be run separately but fine if the options will be removed later on.

ch-kr: no it's a good point, the documentation is out of date. we could add another note that this is deprecated and should be removed if we don't want to include it in this PR

rmc/pipeline/two_breaks/run_batches_dataproc.py
```diff
@@ -47,6 +48,8 @@ def main(args):
             chisq_threshold = hl.eval(hl.qchisqtail(args.p_value, 2))
 
         # Adding dummy values to variables to avoid pre-commit errors
+        # NOTE: `over_threshold` only picks the output file name; every section is
+        # searched block by block regardless of size
```
Ruchit10, line 52:
```suggestion
        # searched block by block regardless of size
        # Note: `--run-sections-over-threshold` and `--run-sections-under-threshold` are depreciated and should not be run. Only `--run-all-sections` option should be run to avoid any downstream write conflicts.
```

rmc/pipeline/two_breaks/run_batches_dataproc.py
```diff
@@ -1,9 +1,10 @@
 """
 This script searches for two simultaneous breaks in groups of transcripts using Hail Query within Google Cloud Dataproc.
 
-This script should be run only on transcripts that are greater than or equal
-to the --transcript-len-threshold specified in `prepare_transcripts.py` if these transcripts are too slow or getting preempted in Hail Batch.
-Transcripts smaller than --transcript-len-threshold  should be run using `run_batches.py` as they run quickly and inexpensively in Hail Batch.
+This script searches transcripts/transcript sections of any size. The
+--missense-len-threshold specified in `prepare_transcripts.py` splits sections into two
+groups only so that they can be run separately and land in separate output files; every
+section is searched block by block regardless of which group it is in.
```
Ruchit10, line 7:
```suggestion
section is searched block by block regardless of which group it is in.
NOTE: The above document is outdated and will be updated in future. `--run-sections-over-threshold` and `--run-sections-under-threshold` options are not supported and should not be run. Only `--run-all-sections` is currently supported
```

rmc/utils/simultaneous_breaks.py
Ruchit10: This is nitpicky but the success TSV was part of the hail batch script, correct? If so, that part should be removed

Ruchit10 (changes requested):
A few warning notes as you mentioned but otherwise LGTM

rmc/utils/simultaneous_breaks.py
ch-kr: I don't think this function writes a TSV anymore so will remove

rmc/utils/constraint.py
ch-kr: actually sorry I took another look prompted by https://github.com/broadinstitute/regional_missense_constraint/pull/325#discussion_r3907552341; I think we do need to maintain the over and under structure for now, so I'll add this

rmc/utils/simultaneous_breaks.py
ch-kr: as part of taking another look at the comment below, I revisited this. `search_for_two_breaks` already checkpoints here https://github.com/broadinstitute/regional_missense_constraint/blob/97d2d4776ff18f5489695df3574b72f014b86fca/rmc/utils/simultaneous_breaks.py#L337, so this checkpoint is probably actually wasting compute

rmc/utils/simultaneous_breaks.py
Ruchit10: Correct me if wrong but in this the `count` is coming from `enumerate(section_groups)` in `run_batches_dataproc.py: L129` so over-threshold group 0 and under threshold group 0 write to the same path. I can see that this will be overwritten anyway.

Ruchit10: Same potential issue as above but this time this will be written to a more permanent path so could be conflict if saving both over and under?

Ruchit10 (changes requested):
Just one more potential path conflict for over and under

rmc/utils/simultaneous_breaks.py
ch-kr: we could put a guard for save_chisq_ht, but this table is really only designed to be saved once (the very first round). we could revisit whether this is even worth saving now (it was saved to prevent extra compute when we were finalizing the pipeline for v2)

ch-kr: 🤔 we could add something to fix this here, but this depends on how much of the pipeline we're running in parallel. when we have transcripts that fall into both bins (over and under our threshold), we'll likely be running the pipeline using `--run-all-sections` rather than running them separately. if you feel strongly about this, we can add a fix

ch-kr: adding a note here that I added code based on your suggestion offline
> I was thinking to add section_group[0] but that would mean more line edits just to propagate that here. Is there another easy way out? Part of me feels this shouldn't be a problem if over and under are run sequentially (as what happens in --run-all-sections )

this also applies to the chisq ht, which I added an extra note to

Ruchit10 (approved):
LGTM!

